PR #8, opened 2018-01-17
LiteEth version bump and initrd update

shenki:


shenki:
Tested on a Arty with the HDMI2USB firmware and Linux

mithro:
This breaks tftp booting the firmware in qemu (IE the ./scripts/build-qemu.sh) right?

shenki:
I expect so

> 15:45 <@mithro> shenki: okay, if you can get things working on the hardware I can fix up the qemu 
>                emulation...

:)

PR #10, opened 2018-01-18
TFTP: Support alternate TFTP Server Port

ewen-naos-nz:
Updates litex submodule pin to include [Alternate TFTP Server Port BIOS support](https://github.com/enjoy-digital/litex/pull/52), where if BIOS is built with TFTP_SERVER_PORT set then it will try TFTP from that port first before falling back to UDP/69.

Update Makefile to start TFTP server on port specified in TFTP_SERVER_PORT if set (defaulting to UDP/69).  If TFTP_SERVER_PORT is >= 1024 then TFTP server is started as user, without using sudo; otherwise TFTP server is started via sudo as before.

`atftpd` path tested; `in.tftpd` path written but not tested.

(Writing this change overlapped with merging of another change around TFTP group, so I've ended up rebase/squashing it into a single commit, but I think I've kept the sense of the overlapping change; I had to do the same thing to be able to run as non-root, otherwise setgid() failed.)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -126,7 +126,7 @@ export TFTP_IPRANGE
 TFTPD_DIR ?= build/tftpd/
 
 # Default TFTP Server Port to IANA well known UDP/69
-TFTP_SERVER_PORT ?= 69
+TFTP_SERVER_PORT ?= 6069
 
 # Couple of Python settings.
 # ---------------------------------
--- a/Makefile
+++ b/Makefile
@@ -125,6 +125,9 @@ TFTP_IPRANGE ?= 192.168.100
 export TFTP_IPRANGE
 TFTPD_DIR ?= build/tftpd/
 
+# Default TFTP Server Port to IANA well known UDP/69
+TFTP_SERVER_PORT ?= 69
+
 # Couple of Python settings.
 # ---------------------------------
 # Turn off Python's hash randomization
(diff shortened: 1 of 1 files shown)


mithro:
@ewen-naos-nz Lets default to the non-privileged port?

Makefile
```diff
@@ -125,6 +125,9 @@ TFTP_IPRANGE ?= 192.168.100
 export TFTP_IPRANGE
 TFTPD_DIR ?= build/tftpd/
 
+# Default TFTP Server Port to IANA well known UDP/69
+TFTP_SERVER_PORT ?= 6069
+
```
mithro: Doesn't this need to be exported?

ewen-naos-nz: The code previously assumed that the user had exported `TFTP_SERVER_PORT` themselves if they wanted a non-root port; based on your change in pull request and discussion it is now exported both in `Makefile` and also in `make env` target that eg, `scripts/build-qemu.sh` uses. This appears to be sufficient that everything sees the default port as UDP/6069, but also allow overriding to use UDP/69 if desired (`TFTP_SERVER_PORT=69`).

Beware that this means that users who built their BIOS a while ago will be trying to TFTP from UDP/69, but their TFTP server will no default to starting on UDP/6069, which may cause additional confusion.... (defaulting to TFTP server on UDP/69 was more backwards compatible).

ewen-naos-nz:
Of note, I still haven't tested `in.tftpd` code path.  It looks sane, but keep an eye out for breakages there.  `atftpd` path is tested.

mithro:
Thank you! This is awesome.

PR #15, opened 2018-01-20
Random collection of fixes.

### LINKED ISSUE #14: third_party/qemu-litex not automatically updated if changed
Sometimes `third_party/qemu_litex` needs updates to match with changes to other litex components.  Eg, 2018-01-20 ethernet updates from `third_party/qemu-litex` in order to successfully TFTP boot arty, otherwise boot hung around the point it started trying to send network packets.

However (a) `third_party/qemu-litex` is not a git submodule, just a git checkout in `scripts/build-qemu.sh` and (b) `scripts/build-qemu.sh` does not automatically fetch newer updates to `third_part/qemu-litex` (or up to a specific "known good" commit, as happens with `git submodule` references when updating), which means it will not automatically be updated.  (Instead the user will need to do `( cd third_party/qemu-litex && git pull)` manually, and find out that's needed some other way.)

mithro:
Bunch of fixes both extracted from @jimmo temporary commits and added while getting LiteEth to boot in QEMU for @shenki.

Fixes #14.

### DIFF
--- a/scripts/build-qemu.sh
+++ b/scripts/build-qemu.sh
@@ -59,7 +59,7 @@ else
 		# Get any new data
 		git fetch $CURRENT_QEMU_REMOTE_NAME
 
-		# Checkout or1k-linux branch it not already on it
+		# Checkout master branch it not already on it
 		if [ "$(git rev-parse --abbrev-ref HEAD)" != "$QEMU_BRANCH" ]; then
 			git checkout $QEMU_BRANCH || \
 				git checkout "$CURRENT_QEMU_REMOTE_NAME/$QEMU_BRANCH" -b $QEMU_BRANCH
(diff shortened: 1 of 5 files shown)


scripts/build-qemu.sh
```diff
@@ -41,6 +45,26 @@ if [ ! -d "$QEMU_SRC_DIR" ]; then
...
+		CURRENT_QEMU_REMOTE_NAME=$(git remote -v | grep fetch | grep "$QEMU_REMOTE_BIT" | sed -e's/\t.*$//')
+		if [ x"$CURRENT_QEMU_REMOTE_NAME" = x ]; then
+			git remote add $QEMU_REMOTE_NAME $QEMU_REMOTE
+			CURRENT_QEMU_REMOTE_NAME=$QEMU_REMOTE_NAME
+		fi
+
+		# Get any new data
+		git fetch $CURRENT_QEMU_REMOTE_NAME
+
+		# Checkout or1k-linux branch it not already on it
```
ewen-naos-nz: From context it looks like this comment needs updating for `scripts/build-qemu.sh`: it appears to be checking out "master" branch here, not `or1k-linux`.  Rest of the change looks plausible though.

PR #16, opened 2018-01-21
tftp: Fix issues with in.tftp

stffrdhrn:
I run on fedora and us in.tftp, it was not working as there were
a few typos in variables and the port argument I dont think does what
it was expected to.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -317,8 +317,8 @@ bios-flash: $(BIOS_FILE) bios-flash-$(PLATFORM)
 ATFTPD=$(shell which atftpd)
 ATFTPD?=/usr/sbin/atftpd
 
-IN_TFTPD=$(shell which in.tfptd)
-IN_TFTPD=?=/usr/sbin/in.tftpd
+IN_TFTPD=$(shell which in.tftpd)
+IN_TFTPD?=/usr/sbin/in.tftpd
 
 tftp: $(FIRMWARE_FILEBASE).bin
 	mkdir -p $(TFTPD_DIR)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -317,8 +317,8 @@ bios-flash: $(BIOS_FILE) bios-flash-$(PLATFORM)
 ATFTPD=$(shell which atftpd)
 ATFTPD?=/usr/sbin/atftpd
 
-IN_TFTPD=$(shell which in.tfptd)
-IN_TFTPD=?=/usr/sbin/in.tftpd
+IN_TFTPD=$(shell which in.tftpd)
+IN_TFTPD?=/usr/sbin/in.tftpd
```
stffrdhrn: Note, I don't think this `?=` does what we expect since IN_TFTPD and ATFTPD vars are not evaluated until later.

ewen-naos-nz: @stffrdhrn: Good catch.  ":=" is needed to force immediate evaluation ("=" is deferred it turns out), and "?=" actually only assigns "if not assigned" not (eg, like shell, perl) if "value is empty".  Reworked in https://github.com/timvideos/litex-buildenv/pull/17 which seems to work better (at least for `atftpd` case in testing here.  Thanks for the heads up.

ewen-naos-nz:
@stffrdhrn AFAICT @mithro ended up merging my change in #17 which contains an equivalent change to the one proposed here (plus a change to try to make `in.tftpd` work on non-default port.  If so, perhaps you might want to close this pull request?  If `in.tftpd` still doesn't work I suspect it'll require a different change than the one to this detection logic.

stffrdhrn:
@ewen-naos-nz Sure I will close.  I think we need some patches though as I find `in.tftpd` cannot work without sudo.

PR #17, opened 2018-01-22
Makefile: tftpd: fix detection logic

ewen-naos-nz:
Rework tftpd detection logic to use immediate Makefile assignment,
and "check for empty" instead of "check for not assigned".  Thanks
to @stffrdhrn for pointing this out:

https://github.com/timvideos/litex-buildenv/pull/16/files#r162814028

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -314,11 +314,15 @@ bios-flash: $(BIOS_FILE) bios-flash-$(PLATFORM)
 # We can run the TFTP server as the user if port >= 1024
 # otherwise we need to run as root using sudo
 
-ATFTPD=$(shell which atftpd)
-ATFTPD?=/usr/sbin/atftpd
+ATFTPD:=$(shell which atftpd)
+ifeq ($(ATFTPD),)
+ATFTPD:=/usr/sbin/atftpd
+endif
 
-IN_TFTPD=$(shell which in.tfptd)
-IN_TFTPD=?=/usr/sbin/in.tftpd
+IN_TFTPD:=$(shell which in.tfptd)
+ifeq ($(IN_TFTPD),)
+IN_TFTPD:=/usr/sbin/in.tftpd
+endif
 
 tftp: $(FIRMWARE_FILEBASE).bin
 	mkdir -p $(TFTPD_DIR)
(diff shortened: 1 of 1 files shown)


ewen-naos-nz:
Also a fair point.  Bother.

My gut feeling is that *so much* happens when we run make now that the `which` is a relatively small overhead.  But if we want to only do the `which` when tftpd targets is run then the solution next best solution I can think of is to push detection down into the shell, inside the `tftpd_start` target.  Eg (off the top of my head):

ATFTPD=$(which atfptd || echo /usr/sbin/atftpd)
IN_TFTPD=$(which in.tftpd || echo /usr/sbin/in.tftpd)

which would then require changing commands to use shell substitution rather than Makefile substitution in the following command lines...

All of which feels like too big a change to make the night before the FPGA Miniconf :-)

(Whereas change in this pull request is, I think, relatively safe give or take overhead of 2 * `which` calls on every `make` run....)

ewen-naos-nz:
Oh, and those `ATFTPD=` and `IN_TFTPD` will either need to use backticks or Makefile escaping to defer expansion until the shell, rather than the Makefile.  All of which feels a bit unreadable.

Maybe we just give up, and just use `?=` (without the `which` variant) and force user to set variable if it is somewhere else.  Or we push this out into the environment, eg, via `scripts/enter-env.sh`, which has the benefit of only ever being done once....

stffrdhrn:
Thanks. I agree with all of the comments above. Moving it out to environment might be good.

Anyway, there is still a typo in the command in.tftpd. my or fixes it.  But I am still having a problem with in.tftpd. 

strace shows it gets the message to download boot.bin. but then it resets the connection.  I'm looking into but don't have much time. It was working before work my last patches around November.

ewen-naos-nz:
@stffrdhrn Sorry to hear that you're still having problems with the `in.tftpd` path.  Unfortunately that's the less tested path (I have `atftpd` installed, and I suspect @mithro probably does too).  In particular I don't know if the "run on port TCP/6069" change actually got properly tested with `in.tftpd`, and maybe it actually does need more than one port available?  (My system has an `/usr/sbin/in.tftpd` binary on it, but it's (a) a link to `atftpd` and (b) takes different arguments -- `atftpd` like ones` so doesn't really help for testing.)

If you do manage to figure out what is needed in the `in.tftpd` path, I'd definitely suggest turning that into a pull request.

PR #19, opened 2018-01-22
Fix some noise and a typo around tftpd setup

thaytan:
in.tftpd still isn't working for me on Fedora 26 with this (in.tftpd[19179]: cannot set groups for user $user) but at least it starts now


thaytan:
Duplicated, I see

Issue #11, opened 2018-01-18
Use an integrated Python based tftp server rather than atftp/in.tftpd on the system.


mithro:
It might still be worth providing, eg, Python-based TFTP server to reduce need to install system dependencies that auto-start daemons. (But I suspect some users will be forced to interact with their firewall config anyway.)

ewen-naos-nz:
(Copying from issue #6 for ease of reference):

tftpy code appears to be at https://github.com/msoulier/tftpy now. Last update 2017 (cf sf.net page last updated around 2010). Seems to have a listenport option on the TFTPServer class, so at least could be given a simple frontend to listen on a different port.

py3tftp (https://github.com/sirMackk/py3tftp) is server only, takes port on command line, and defaults to UDP/9069.

Potentially either one would do as a "simple TFTP Server run as user".

PR #21, opened 2018-02-06
tftpd: Fix in.tftpd alternate port syntax

ewen-naos-nz:
This fixes the most obvious issue with running `in.tftpd` as a non-root user, by getting the syntax for the listening port correct.  Unfortunately at least when run on Ubuntu 16.04 the next result is:

    Feb  7 11:39:20 parthenon in.tftpd[1690]: cannot set groups for user ewen
    Feb  7 11:39:22 parthenon in.tftpd[1691]: cannot set groups for user ewen
    Feb  7 11:39:25 parthenon in.tftpd[1694]: cannot set groups for user ewen
    Feb  7 11:39:27 parthenon in.tftpd[1695]: cannot set groups for user ewen
    Feb  7 11:39:30 parthenon in.tftpd[1699]: cannot set groups for user ewen

(one for each TFTP REQ attempt).  I've found references to that going back 10 years, so I'm not 100% sure it will be possible to run `in.tftpd` as a non-`root` non-`nobody` (default user it changes to) user.  Needs more investigation, but pushing WIP pull request as this is clearly closer than previous attempt.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -343,6 +343,10 @@ ATFTPD:=/usr/sbin/atftpd
 endif
 
 # Requires HPA in.tftpd not traditional BSD in.tftpd
+# Unfortunately in.tftpd seems to require root always
+# even if run as current user, otherwise it reports
+# "cannot set groups for user $USER"
+#
 IN_TFTPD:=$(shell which in.tfptd)
 ifeq ($(IN_TFTPD),)
 IN_TFTPD:=/usr/sbin/in.tftpd
@@ -358,7 +362,11 @@ tftpd_stop:
 		sudo true; \
 		sudo killall atftpd || sudo killall in.tftpd || true; \
 	else \
-		killall atftpd || killall in.tftpd || true; \
+		killall atftpd || \
+		if ps axuwww | grep -v grep | \
+			grep "[i]n.tftpd" >/dev/null 2>&1; then \
+				sudo killall in.tftpd; \
+		fi || true; \
 	fi
 
 tftpd_start:
@@ -379,7 +387,9 @@ tftpd_start:
 		if [ $(TFTP_SERVER_PORT) -lt 1024 ]; then \
 			sudo "$(IN_TFTPD)" --verbose --listen --address $(TFTP_IPRANGE).100:$(TFTP_SERVER_PORT)  --user $(shell whoami) -s $(TFTPD_DIR) & \
 		else \
-			"$(IN_TFTPD)" --verbose --listen --address $(TFTP_IPRANGE).100:$(TFTP_SERVER_PORT) --user $(shell whoami) -s $(TFTPD_DIR) & \
+			echo "Root required to run in.tftpd, will use sudo"; \
+			sudo true; \
+			sudo "$(IN_TFTPD)" --verbose --listen --address $(TFTP_IPRANGE).100:$(TFTP_SERVER_PORT) --user $(shell whoami) -s $(TFTPD_DIR) & \
 		fi \
 	else \
 		echo "Cannot find an appropriate tftpd binary to launch the server."; \
--- a/Makefile
+++ b/Makefile
@@ -342,6 +342,7 @@ ifeq ($(ATFTPD),)
 ATFTPD:=/usr/sbin/atftpd
 endif
 
+# Requires HPA in.tftpd not traditional BSD in.tftpd
 IN_TFTPD:=$(shell which in.tfptd)
 ifeq ($(IN_TFTPD),)
 IN_TFTPD:=/usr/sbin/in.tftpd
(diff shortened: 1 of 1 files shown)


ewen-naos-nz:
If `--user $(shell whoami)` is omitted, `in.tftpd` will default to `--user nobody`, so that doesn't help.

Digging around for causes online finds a 14 year old Debian bug discussion, eg:

https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=245017#44

which indicates `in.tftpd` calls `setgroups32()`, which can only be run as `root` (well technically only with `CAP_SETGID` but usually you get that via, eg, `sudo`), even to change groups to the same groups that are loaded.  Interestingly [Debian closed that bug in 2009](https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=245017#82) with the commit log comment:

<pre>
tftp-hpa (5.0-2) unstable; urgency=low

  * Adding forgotten bug numbers to previous changelog entry.
  * Now running always as unprivileged user (Closes: #245017).

 -- Daniel Baumann <mail@daniel-baumann.ch>  Mon, 20 Jul 2009 01:27:27 +0200
</pre>

from which I conclude we should probably do the same -- ie, always run `in.tftpd` as root, no matter the port number.  I haven't found a way around it accessible from the command line.  So `atftpd` is probably a better choice to recommend until we switch to a Python based tftpd server (see issue https://github.com/timvideos/litex-buildenv/issues/11).

Ewen

ewen-naos-nz:
With change to always run `in.tftpd` with sudo, it now appears to work on `UDP/6069`:

    export TFTP_SERVER_PORT=6069
    make tftpd_stop
    make tftpd_start
    ps axuw | grep [t]ftpd
    make firmware-connect

gives:

<pre>
(LX P=arty C=or1k.linux F=linux R=in-tftpd-fix) ewen@parthenon:/src/fpga-minicon
f/litex-buildenv$ ps axuw | grep [t]ftpd
root     17489  0.0  0.0  15180   152 ?        Ss   19:04   0:00 /usr/sbin/in.tftpd --verbose --listen --address 192.168.100.100:6069 --user ewen -s build/tftpd/
(LX P=arty C=or1k.linux F=linux R=in-tftpd-fix) ewen@parthenon:/src/fpga-minicon
f/litex-buildenv$ make firmware-connect
flterm --port=/dev/ttyUSB1 --speed=115200
[FLTERM] Starting...
        __   _ __      _  __
       / /  (_) /____ | |/_/
      / /__/ / __/ -_)>  <
     /____/_/\__/\__/_/|_|
      SoC BIOS / CPU:OR1K
(c) Copyright 2012-2018 Enjoy-Digital
(c) Copyright 2007-2018 M-Labs Limited
Built Feb  7 2018 18:04:07

BIOS CRC passed (1f086202)
Initializing SDRAM...
Memtest OK
Booting from serial...
Press Q or ESC to abort boot completely.
sL5DdSMmkekro
Timeout
Booting from network...
Local IP : 192.168.100.50
Remote IP: 192.168.100.100
Fetching from: UDP/6069
Successfully downloaded 5993804 bytes from boot.bin over TFTP
Unable to download cmdline.txt over TFTP
No command line parameters found
Unable to download initrd.bin over TFTP
No initial ramdisk found
Executing booted program at 0x40000000
</pre>

on real hardware, and also in `qemu`.

I've also retested with `atftpd` and it appears that can still avoid using root (even in `tftpd_stop` which required some shell magic to avoid in the case where `atftpd` is not running).

Ewen

ewen-naos-nz:
Removed WIP, as this now appears to work for me.

@stffrdhrn Could you test this version?  If it works (sorry, appears `sudo` is always required with `in.tftpd` :-( ) then @mithro could merge this version which appears to be the best we can do for `in.tftpd`.  Thanks.

Ewen

mithro:
Maybe we should bite the bullet and use a python tftpd server?

ewen-naos-nz:
@mithro I think it'd definitely make things easier.  But I'd also be inclined to merge this pull request as it seems to me clearly better for `in.tftpd` than what is there now.

Ewen

Makefile
```diff
@@ -342,6 +342,11 @@ ifeq ($(ATFTPD),)
 ATFTPD:=/usr/sbin/atftpd
 endif
 
+# Requires HPA in.tftpd not traditional BSD in.tftpd
+# Unfortunately in.tftpd seems to require root always
+# even if run as current user, otherwise it reports
+# "cannot set groups for user $USER"
+#
 IN_TFTPD:=$(shell which in.tfptd)
```
stffrdhrn: Here `tftpd` is spelled wrong

Makefile
```diff
@@ -376,9 +385,11 @@ tftpd_start:
 	elif [ -x "$(IN_TFTPD)" ]; then \
 		echo "Starting in.tftpd"; \
 		if [ $(TFTP_SERVER_PORT) -lt 1024 ]; then \
```
stffrdhrn: Do we need this if/else if its always sudo?

Makefile
mithro: I also think that `in.tftpd` is not normally in non-root users path.

Makefile
ewen-naos-nz: The if/else only exists now so that we can display a "root required" warning in the "high-port, using `in.tftpd`" case, as the "root required" warning at the top is only displayed for low numbered ports.  I did look at trying to rejig the top test to only skip it if (a) `atftpd` and (b) high numbered port... but it started becoming rather more complicated to follow.

Makefile
ewen-naos-nz: @stffrdhrn Pushed new patch (finally!) fixing the typo in `$(which in.tftpd)`, which I thought I'd fixed previously, but apparently never in a branch I pushed :-(   (You've pointed it out several times!)

@mithro `in.tftpd` is in my path :-)  But yes, I'm a bit unusual.  There's a hard coded default assumption in case the `$(which in.tftpd)` wasn't matched, which is what was working before.  I think it's worth trying the `$(which ...)` in case someone has installed it as a local file, rather than a system package.

ewen-naos-nz:
@stffrdhrn @mithro I've pushed a third commit fixing the type in `$(which ...)` (which @stffrdhrn pointed out multiple times :-( ).  So hopefully *this* is the best we can do with `in.tftpd` :-)

mithro:
LGTM then!

PR #24, opened 2018-02-10
Makefile: make FIRMWARE=none work, Mimas v2 gateware-flash redirected to include BIOS

ewen-naos-nz:
If `FIRMWARE=none ` then the firmware will be omitted in `make image-flash`, but previously it would still try to build a `none.fbi` firmware image to include.  Move `.fbi` dependency into a variable so that we can replace it with a blank string if `FIRMWARE=none`.  Appears to work as expected for both `FIRMWARE=micropython` and `FIRMWARE=none`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -178,11 +178,13 @@ third_party/%/.git: .gitmodules
 # --------------------------------------
 ifeq ($(FIRMWARE),none)
 OVERRIDE_FIRMWARE=--override-firmware=none
+FIRMWARE_FBI=
 else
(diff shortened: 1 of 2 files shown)


ewen-naos-nz:
@mithro As discussed on IRC also added diversion from "make gateware-flash" just loading gateware (and no BIOS, and thus being not very useful) on Mimas v2, to have it load `image-gateware+bios+none.bin` instead, which means "`make gateware-flash`" and "`make firmware-load`" can actually work without having to flash the application firmware image.

Ewen

mithro:
I like what you have done here!

PR #25, opened 2018-02-19
Fixes to PROMPT_COMMAND

tsukasa-au:
Two misc fixes:
 - to remove errors when in.tftpd is not installed, and
 - prepend to PROMPT_COMMAND, rather than overwriting its contents.


mithro:
LGTM!

PR #35, opened 2018-08-27
Add FULL_PLATFORM to Makefile 'env' command

MattSegal:


MattSegal:
The `/scripts/download-prebuild.sh` command requires this envar. It breaks without it.

PR #45, opened 2018-08-28
Makefile: pass CPU_VARIANT through to firmware build

ewenmcneill:
With this change we can build `CPU=lm32` `CPU_VARIANT=minimal`, and boot MicroPython on the Mimas V2 and the Arty.  It appears with the `-Ot cpu_variant minimal` the firmware does not get built in the right directory, which causes building MicroPython to fail (and possibly other things).

Boot on the Mimas V2 is *very* slow with this variant (possibly several timeouts).  Boot on the Arty seems relatively quick.  I've not attempted to debug the cause of this, but this trivial fix appears necessary for `CPU_VARIANT=minimal` to build.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -91,7 +91,7 @@ ifeq ($(CPU_VARIANT),)
 FULL_CPU = $(CPU)
 else
 FULL_CPU = $(CPU).$(CPU_VARIANT)
-MAKE_LITEX_EXTRA_CMDLINE += -Ot cpu_variant $(CPU_VARIANT)
+MAKE_LITEX_EXTRA_CMDLINE += --cpu-variant=$(CPU_VARIANT)
 endif
 
 # Include platform specific targets


mithro:
@ewenmcneill minimal builds the lm32 without any cache.

The Arty executes the firmware from DDR memory which has its own cache.
The MimasV2 executes from SPI flash which is very slow.

We can play with the lm32 configuration in the future I think?

ewenmcneill:
@mithro Cool, that explains the massive speed difference I saw.

Yes, definitely merge this change.  It's needed to compile with `CPU_VARIANT=minimal`, and does work.  But we may want to consider adding some more variants with different subsets of `lm32` features later.

Ewen

enjoy-digital:
@ewenmcneill: strange for the firmware that is not build in the right directory.

The minimal variant was created for very simple design that need to reduce ressource usage to minimum (ex tinyfpga-soc), or where it's more convenient to use a cpu than a hardcoded FSM. Feel free to suggest others variants if you need others. (You just need to customize the lm32_config.v file).

ewenmcneill:
@enjoy-digital The firmware is built in the right directory now -- it looks like the `Makefile` and `make.py` had different ideas of how they wanted to receive the `CPU_VARIANT` setting which was previously resulting in `make.py` ignoring it, but everything else in the Makefile using it.

I suspect at some point we'll want a "small" variant or similar, which has a little more turned on to be useful on boards that don't have fast "main RAM" (eg, DDR3 with at least a row cache on-chip), but still leaves more space for non-CPU things on the smaller FPGA chips.  But that can be done when we have a specific use case.  (@mithro and I wanted to test "minimal" today just to have options when trying to target the iCE 40 TinyFPGA B2/B2 boards, which even with the largest iCE 40 are still relatively small FPGAs...)

Ewen

ewenmcneill:
@enjoy-digital For the record I wanted to echo what [cr1901_modern said on IRC](https://logs.timvideos.us/%23timvideos/%23timvideos.2018-08-28.log.html) (also cut'n'paste below) -- that without an instruction cache (i-cache) performance will be *very* slow on systems without fast main RAM/ROM. 

In particular I have a feeling that we're doing execute-in-place out of the SPI Flash on the Mimas v2 (for space constraint reasons -- I think @mithro wants to do the same thing on the iCE 40 TinyFPGA boards).  Without an instruction cache I suspect it'll be fetching *every* instruction all the way from the SPI Flash, which I think has fairly high latency (compared with cache/DDR3 RAM, etc) for single byte fetches.  That might well explain the *very* slow boot performance on the Mimas V2 with the `lm32` `minimal` profile -- it took in the order of a minute to boot through the BIOS and into MicroPython, from flash (ie, no serial load), with the `lm32` `minimal`.  IIRC the "memory test" was one of the slowest bits, and that's presumably an inner loop which would benefit a lot from even a small i-cache.  (IIRC the Arty boots from in-FPGA ROM and/or ROM copied into main RAM; and it booted fast enough with `lm32` `minimal` that it wasn't really noticeably slower than the regular `lm32` just looking at the serial console.)

I have a feeling that even an i-cache that's the size of modern DRAM row buffers (128 bytes?) would make a huge performance impact on high latency backing ROM.  (d-cache is probably a lower priority, as the data accesses are generally both less frequent and also usually hitting lower latency main RAM anyway.)

So possibly that'd be the first thing to put in a "tiny" or "small" profile, which was one step above the minimum required for the processor to boot -- a small i-cache.

Ewen

<pre>

cr1901_modern | mithro: Does this include with spiflash?
cr1901_modern | Also please note that a truly minimal lm32 doesn't have mul/div/multishift insns | 
cr1901_modern | which Idk if litex accounts for when generating Makefile flags
cr1901_modern | >doesn't have mul/div/multishift insns
cr1901_modern | Ignore this, just actually read the source. I still think the minimal lm32 should have an i-cache though 
cr1901_modern | cc: _florent_
cr1901_modern | The  idea is that designs using tinyfpga and ice40 SoCs will typically be  reading from an external flash (assuming I ever get the damn thing to  work)
cr1901_modern | if  you don't have a cache, then there's no reason to use a pipelined CPU  b/c the pipeline will never be filled due to throughput from SPIflash
</pre>

enjoy-digital:
@ewenmcneill @cr1901, thanks. In my case the "minimal" variant was used with instructions in ROM, but yes we can create another variant ("tiny"?) that would better fit your needs.

PR #52, opened 2018-09-08
QEMU_REMOTE, or1k, tabs and CFLAGS

CarlFK:
re-did my work, tried to isolate changes, almost got it perfect but forgot that hitting tab inserts 4 spaces.

there were way more tabs than lines with 4 spaces, so I picked tab to be what indents. 
I think there are more files in the scripts dir that are a mix.


scripts/build-qemu.sh
```diff
@@ -34,38 +34,38 @@ set -x
...
-		if [ "$(git rev-parse --abbrev-ref HEAD)" != "$QEMU_BRANCH" ]; then
-			git checkout $QEMU_BRANCH || \
-				git checkout "$CURRENT_QEMU_REMOTE_NAME/$QEMU_BRANCH" -b $QEMU_BRANCH
-		fi
-	)
-fi
+    # Get any new data
+    git fetch $CURRENT_QEMU_REMOTE_NAME $QEMU_BRANCH
+
+	# Checkout master branch if not already on it
```
mithro: Indenting here looks wrong?

PR #55, opened 2018-09-11
Add TinyFPGA BX Platform

cr1901:
Because the stock `litex-buildenv` doesn't fit in TinyFPGA's SRAM, I have opted for a stub firmware for now. The gateware begins at `0x28000` in SPI flash, while the BIOS begins at `0x50000`, and the firmware at `0x58008` (allowing 2 32-bit words for length and checksum).

I have tested that a proper image is compiled and flashes correctly on Windows; I have only tested the former on Linux after sourcing `./scripts/enter-env.sh`.

The following environment vars _must_ be defined before sourcing `./scripts/enter-env.sh`:

```
PLATFORM=tinyfpga_bx TARGET=base CPU=lm32 CPU_VARIANT=lite FIRMWARE=stub
```

~I would not try routing this design with `arachne-pnr`: the I/O pads used on the BX seem to be spread out sufficiently far that `arachne-pnr` has trouble consistently routing a working design.~

EDIT: `arachne-pnr` is unsupported. It works okay for tinyfpga B2, but tends to choke on BX.

You will need the following applications installed that are likely not provided by `conda`:
* https://github.com/tinyfpga/TinyFPGA-Bootloader
* https://github.com/YosysHQ/nextpnr


firmware/pattern.c
```diff
@@ -16,6 +16,7 @@ unsigned int pattern_framebuffer_base(void) {
 	return FRAMEBUFFER_BASE_PATTERN;
 }
 
+#ifdef MAIN_RAM_BASE
```
mithro: I'm not sure why this define is here? Being `static const` should mean it doesn't end up in the SRAM?

firmware/pattern.c
```diff
@@ -26,7 +27,9 @@ static const unsigned int color_bar[8] = {
 	YCBCR422_BLUE,
 	YCBCR422_BLACK
 };
+#endif
 
+#ifdef MAIN_RAM_BASE
```
mithro: I'm not sure why this define is here? Being `static const` should mean it doesn't end up in the SRAM?

targets/tinyfpga_bx/base.py
```diff
@@ -0,0 +1,127 @@
...
+from litex.build.generic_platform import Pins, Subsignal, IOStandard
+from litex.soc.integration.soc_core import *
+from litex.soc.integration.builder import *
+
+from gateware import info
+from gateware import spi_flash
+
+from targets.utils import csr_map_update
+
+
```
mithro: This seems duplicate with the `platform/tinyfpga_bx.py` file?

targets/tinyfpga_bx/base.py
```diff
@@ -0,0 +1,127 @@
...
+        if 'integrated_sram_size' not in kwargs:
+            kwargs['integrated_sram_size']=0x2800
+
+        # FIXME: Force either lite or minimal variants of CPUs; full is too big.
+
+        platform.add_extension(serial)
+        platform.add_extension(reset)
+        clk_freq = int(16e6)
+
+        # Extra 0x28000 is due to bootloader bitstream.
```
mithro: `BOOTLOADER_SIZE` would probably be a good define to put on the platform object...

Magic numbers bad, m'kay? :-P

targets/tinyfpga_bx/base.py
```diff
@@ -0,0 +1,127 @@
...
+            # Leave a grace area- possible one-by-off bug in add_memory_region?
+            # Possible fix: addr < origin + length - 1
+            platform.spiflash_total_size - (self.flash_boot_address - self.mem_map["spiflash"]) - 0x100)
+
+        # Disable USB activity until we switch to a USB UART.
+        self.comb += [platform.request("usb").pullup.eq(0)]
+
+
+        # Annotate build scripts with required options.
+        # Seed 2 in general works better than seed 1.
```
mithro: I think this is just luck.....

targets/tinyfpga_bx/base.py
```diff
@@ -0,0 +1,127 @@
...
+        platform.toolchain.build_template[1] = \
```
mithro: These should most certainly be part of litex / migen in some way in the future?

targets/tinyfpga_bx/base.py
cr1901: I agree! We shouldn't be using `arachne` for this. I just put it there in case someone tries.

firmware/pattern.c
cr1901: Without this gate (and without `FIRMWARE=stub`), I get the following error:

```
 CC       pattern.o
C:/msys64/home/William/prj/Consult/timvideos/litex-buildenv/firmware/pattern.c:161:13: warning: 'pattern_draw_text' defined but not used [-Wunused-function]
 static void pattern_draw_text(int x, int y, char *ptr) {
             ^~~~~~~~~~~~~~~~~
C:/msys64/home/William/prj/Consult/timvideos/litex-buildenv/firmware/pattern.c:133:12: warning: 'inc_color' defined but not used [-Wunused-function]
 static int inc_color(int color) {
            ^~~~~~~~~
C:/msys64/home/William/prj/Consult/timvideos/litex-buildenv/firmware/pattern.c:33:28: error: 'font5x7' defined but not used [-Werror=unused-const-variable=]
 static const unsigned char font5x7[] = {
                            ^~~~~~~
C:/msys64/home/William/prj/Consult/timvideos/litex-buildenv/firmware/pattern.c:20:27: error: 'color_bar' defined but not used [-Werror=unused-const-variable=]
 static const unsigned int color_bar[8] = {
                           ^~~~~~~~~
cc1.exe: all warnings being treated as errors
make[1]: *** [C:/msys64/home/William/prj/Consult/timvideos/litex-buildenv/firmware/Makefile:106: pattern.o] Error 1
make[1]: Leaving directory '/home/William/prj/Consult/timvideos/litex-buildenv/build/tinyfpga_bx_base_lm32.lite/software/firmware'
Traceback (most recent call last):
  File "./make.py", line 151, in <module>
    main()
  File "./make.py", line 135, in main
    vns = builder.build(**dict(args.build_option))
  File "C:/msys64/home/william/prj/projects/litex\litex\soc\integration\builder.py", line 160, in build
    self._generate_software(not self.soc.integrated_rom_initialized)
  File "C:/msys64/home/william/prj/projects/litex\litex\soc\integration\builder.py", line 134, in _generate_software
    subprocess.check_call(["make", "-C", dst_dir, "-f", makefile])
  File "C:/msys64/mingw64/lib/python3.6\subprocess.py", line 291, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['make', '-C', 'C:/msys64/home/William/prj/Consult/timvideos/litex-buildenv/build/tinyfpga_bx_base_lm32.lite/software/firmware', '-f', 'C:/msys64/home/William/prj/Consult/timvideos/litex-buildenv/firmware/Makefile']' returned non-zero exit status 2.
```

I added the gate, based on the following [comment](https://github.com/timvideos/litex-buildenv/blob/3f691e817cb7be802b760646ff84a572ee8e28d9/firmware/pattern.c#L157):

```
/* FIXME: Framebuffer Should not even be compiled if no MAIN RAM */
```

So it's a band-aid :P, but what do you think we should do?

targets/tinyfpga_bx/base.py
cr1901: Based on IRC comments:
```
(6:59:33 PM) mithro: cr1901_modern: I mean there should just be switches which modify the template in some way rather than having to override whole template 
```

It was a conscious design [decision](https://github.com/m-labs/migen/pull/84#discussion_r149899202) to move away from what you're suggesting in Migen. I think you should talk to @jordens and see what he thinks; the `icestorm` backend at present _is_ used to test ideas. I don't know what such a feature you're suggesting would look like short of an explosion of new arguments to the `plat.build` method.

targets/tinyfpga_bx/base.py
cr1901: I don't see `csr_map_update` imported anywhere in `platform/tinyfpga_bx/py`?

firmware/pattern.c
ewenmcneill: @cr1901 @mithro It looks like the *correct* fix would be to have a BIOS build target that does not even build/link `pattern.c`.  AFAICT that's something that really only makes sense on video output boards (Opsis, etc).  Eg, maybe it shouldn't be built for `TARGET=BASE`?

Ewen

targets/tinyfpga_bx/base.py
mithro: I think this comment was actually attached to code you removed. Closing.

targets/tinyfpga_bx/base.py
mithro: Okay, lets worry about that later.

targets/tinyfpga_bx/base.py
mithro: I'm pretty sure you are just getting lucky with Seed 2 verse Seed 1....

firmware/pattern.c
mithro: Looking into this deeper, it seems that in C, these values are not optimized away (compared to C++ were they would be). From https://nullprogram.com/blog/2016/07/25/;
> In general, const declarations can’t help the optimizer, but const definitions can.

ewenmcneill: FWIW, I've found my commit (from pre-LCA2018 hack fest) which I think included the `FIXME`, etc) when we were trying to get the BIOS to fit into a smaller system (that one ultimately didn't work because the Spartan 3 FPGA part was *tiny* and not even a minimal CPU would fit).   

Looks like [that change eventually got merged some other way](https://github.com/timvideos/litex-buildenv/commit/87b46c0248d82cc1211464ba582da196e802bb1b) (maybe via someone else trying to get things working).  But [comment on commit explains the rationale for the first hack](  https://github.com/ewenmcneill/litex-buildenv/commit/87b46c0248d82cc1211464ba582da196e802bb1b), which I hadn't intended to be merged in that form...  (I've still got the branch sitting around as I had assumed it didn't get merged!)

Ewen

targets/tinyfpga_bx/base.py
cr1901: My suggestion is to leave it there for now, but at some point in the near future deprecate `arachne-pnr` for this platform. For whatever reason, it is more difficult to route on BX than it is on B2.

To answer your q on IRC, BX and B2 have the same form factor. B2 routes the USB+/-, clk, and SPI flash to some pins, while BX leaves the same pins free for GPIO. This results in pretty different UCFs, which is why I separated the platforms in LiteX/Migen in the first place. I suspect BX's UCF is such that `arachne-pnr` just genuinely has trouble routing while satisfying constraints.

firmware/pattern.c
cr1901: `TARGET` is available to the `Makefile`. Since by convention, `base` systems don't have video, it's probably alright to modify the firmware `Makefile` to only append `pattern.c` if we are compiling a design more comprehensive than `TARGET=base`.

ewenmcneill: @cr1901 @mithro I've created https://github.com/timvideos/litex-buildenv/issues/59 to track this desire to build a minimised version of the BIOS for `TARGET=base`.  I remember us talking about it in January too, but it looks like the kludgey "`#ifdef` some things out to find out if it'll even fit" version got merged as-is (presumably along with some other changes).  I'd suggest taking this pull request including the extra kludges in `firmware/pattern.c` and working on `Makefile` changes for a more minimal BIOS separately under https://github.com/timvideos/litex-buildenv/issues/59.  

Ewen

mithro:
```
The following builds succeeded
=============================================
tinyfpga_bx base lm32.lite stub
```

Issue #56, opened 2018-09-11
current qemu: does not support the option 'serial'


CarlFK:
```
export QEMU_REMOTE=https://github.com/stffrdhrn/qemu.git
export QEMU_BRANCH=litex
(LX P=arty C=or1k) $ ./scripts/build-qemu.sh

...

cp build/arty_net_or1k//software/firmware/firmware.bin build/tftpd//boot.bin
+ EXTRA_ARGS+=("-gdb tcp::10001")
++ grep spiflash_model platforms/arty.py
++ sed '-es/[^"]*"//' '-es/".*$//'
+ SPIFLASH_MODEL=n25q128a13
+ echo n25q128a13
n25q128a13
+ build/arty_net_or1k//qemu/or1k-softmmu/qemu-system-or1k -M litex -nographic -nodefaults -monitor telnet::10000,server,nowait -serial stdio -bios build/arty_net_or1k//software/bios/bios.bin -drive if=mtd,format=qcow2,file=build/arty_net_or1k//qemu.qcow2,serial=n25q128a13 -net nic -net tap,ifname=tap0,script=no,downscript=no -gdb tcp::10001
qemu-system-or1k: -drive if=mtd,format=qcow2,file=build/arty_net_or1k//qemu.qcow2,serial=n25q128a13: Block format 'qcow2' does not support the option 'serial'
```

stffrdhrn:
This is fixed. I don't think I have access to close this bug though?

PR #57, opened 2018-09-13
qemu: Fix spiflash args as per deprecated args

stffrdhrn:
The arg available to specify serial for -drive was to be used for
specifying a drive serial number.  However, we we using it to specify
the flash driver.  This was a hack.

Introduce a driver option to be able to specify this on litex_ssi.  This
will then be read during ssi bus / drive initialization to initialize
the requested guest hardware.

By default we fall back to m25p80.

Example usage:
  -drive if=mtd,format=qcow2,file=build/arty_net_or1k//qemu.qcow2
  -global litex_ssi.spiflash=mx25l1606e

This depends on qemu patches to be used, over at: 
  https://github.com/stffrdhrn/qemu/tree/litex

If the patches are not available it the argument will just be ignored.


stffrdhrn:
This should fix #56

PR #58, opened 2018-09-13
qemu: Allow qemu build to get upstream qemu after changes

stffrdhrn:
Currently the qemu build will checkout the requested qemu branch if we
are not on it.  However, if we are on the correct branch it will not do
anything.  This is usually fine, but if upstream has changes it means
those changes will be be able to get picked up.

Add a 'git reset' to allow picking up of the upstream changes.

Also, currently the QEMU_REMOTE_NAME is always the same, that means if
we override the remote we will not really be able to add the new remote.
To enable this create the QEMU_REMOTE_NAME based on the remote url.

### DIFF
--- a/scripts/build-qemu.sh
+++ b/scripts/build-qemu.sh
@@ -35,8 +35,10 @@ set -e
 
 QEMU_REMOTE="${QEMU_REMOTE:-https://github.com/timvideos/qemu-litex.git}"
 QEMU_BRANCH=${QEMU_BRANCH:-master}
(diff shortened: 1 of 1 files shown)


stffrdhrn:
There are some problems with this, i.e. we end of getting 2 remotes pointing the the timvideos/qemu-litex.  This causes it to think we are on both origin and timvideos-qemu-litex and then it breaks.  I will need to rethink how to make this nice.

PR #87, opened 2018-10-07
scripts: Use GCC 8.2.0 on or1k too.

mithro:


mithro:
@stffrdhrn - This pull request changes our or1k build to use your GCC changes. It currently fails with `or1k-elf-gcc: error: unrecognized command line option '-mror'; did you mean '-Werror'` which I assume is because you don't have the `-mror` option yet?

stffrdhrn:
I have added mror recently. Let me check if you got it.

https://github.com/stffrdhrn/gcc/blob/or1k-port-2b/gcc/config/or1k/or1k.opt

mithro:
I'm using https://github.com/mithro/gcc which is your `or1k-port` branch rebased onto upstream gcc-8.2

stffrdhrn:
Right, that didn't have mror.

stffrdhrn:
By the way, I try to create a new branch for every patch series I send to
the mailing list for review.  That's why there are so many branches.

mithro:
Is there a way to have a branch which is just considered the "latest"?

mithro:
@stffrdhrn I believe your latest patchset means things compiles fine. Need to test it actually works. The build is currently running here -> https://travis-ci.org/timvideos/litex-buildenv/builds/439432758

PR #89, opened 2018-10-12
Firmware CR-NL

stffrdhrn:
Support different combinations of CR/NL as is now down in the litex bios.

### DIFF
--- a/firmware/ci.c
+++ b/firmware/ci.c
@@ -273,9 +273,9 @@ static char *readstr(void)
 					break;
 				case 0x07:
 					break;
-				case '\n':
-					break;
 				case '\r':
+					break;
+				case '\n':
 					s[ptr] = 0x00;
 					wputs("");
 					ptr = 0;
--- a/firmware/ci.c
+++ b/firmware/ci.c
@@ -236,7 +236,7 @@ static char *readstr(void)
 	if(telnet_active) {
 		if(telnet_readchar_nonblock()) {
 			c[0] = telnet_readchar();
-			c[1] = 0;
+			c[1] = '\0';
 			switch(c[0]) {
 				case 0x7f:
 				case 0x08:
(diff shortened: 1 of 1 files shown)


firmware/ci.c
```diff
@@ -231,6 +231,7 @@ static char *readstr(void)
 	char c[2];
 	static char s[128]; // Needs to fit a full mode line ~100 chars.
 	static int ptr = 0;
+	static char skip = 0;
```
mithro: Should this technically be `'\0'` ?

mithro: Maybe something like `skip_if_next` is a better name? 
We only skip the value if it is exactly the *next* value to be received right? IE `\ni\r` wouldn't hit the skip?

firmware/ci.c
```diff
@@ -263,6 +264,14 @@ static char *readstr(void)
 		if(readchar_nonblock()) {
 			c[0] = readchar();
 			c[1] = 0;
+
+			/* Check if we have anything to skip.  */
```
mithro: I don't really like the reusage of `skip` variable here...

Maybe
```c
bool should_skip = (c[0] == skip);
skip = '\0';
if (should_skip)
  return NULL;
```

or

```c
if (c[0] == skip) {
  skip = '\0';
  return NULL;
} else {
  skip = '\0';
}
```

mithro:
Concept seems fine! Just some minor nits around the style....

firmware/ci.c
stffrdhrn: Yes, I was just following the style.  I can update other places they are assigning char to `0` to `'\0'`

stffrdhrn: Yes, only next.  I can change the name something like that.  Again, I was just following the naming from litex.

firmware/ci.c
stffrdhrn: I had it like the second case before.  I was just trying to be clever. I will revise.

stffrdhrn:
I will clean them up.  The CI failure doesn't look related, I couldn't see what it was in the log though.

mithro:
LGTM!

PR #90, opened 2018-10-15
ICE40UP5K-B-EVN Support

cr1901:
This PR adds preliminary support for ICE40 UltraPlus Family members with 5280 LUTs. The design is a tight fit, even with the `minimal` variant (and this is _without_ features like `ControlAndStatus` or `spi`/`i2c`). 

Both the `minimal` and `lite` variants of the `lm32` CPU are tested pending further optimizations; I _really_ want the cache of the `lite` variant, but it might not be possible to reduce resource usage much further.

Currently TODO:
* RGB LEDs (using the built-in primitive?)
* _Possibly_ SPI/I2C IP (using the built-in cores?)?

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -83,6 +83,9 @@ jobs:
     env: C=lm32.lite P=ice40_up5k_b_evn F=stub T="base"
 
   - stage: Targets
+    env: C=lm32.minimal P=ice40_up5k_b_evn F=stub T="base"
+
+  - stage: Targets
     env: C=lm32 P=mimasv2      T="base"
 
   - stage: Targets
--- a/.travis.yml
+++ b/.travis.yml
@@ -80,6 +80,9 @@ jobs:
     env: C=lm32.lite P=ice40_hx8k_b_evn F=stub T="base"
 
   - stage: Targets
+    env: C=lm32.lite P=ice40_up5k_b_evn F=stub T="base"
+
+  - stage: Targets
     env: C=lm32 P=mimasv2      T="base"
 
   - stage: Targets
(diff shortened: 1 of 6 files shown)


mithro:
Just two minor comments, otherwise LGTM!

.travis.yml
```diff
@@ -79,6 +79,12 @@ jobs:
   - stage: Targets
     env: C=lm32.lite P=ice40_hx8k_b_evn F=stub T="base"
 
+  - stage: Targets
+    env: C=lm32.lite P=ice40_up5k_b_evn F=stub T="base"
+
+  - stage: Targets
+    env: C=lm32.minimal P=ice40_up5k_b_evn F=stub T="base"
```
mithro: Why is there a `lm32.lite` and `lm32.minimal`?

targets/ice40_up5k_b_evn/Makefile.mk
```diff
@@ -0,0 +1,66 @@
...
+	@false
+
+# As with Mimasv2, if the user asks to flash the gateware only, the BIOS must
+# be sent as well (because the BIOS is too big to fit into the bitstream).
+GATEWARE_BIOS_FILE = $(TARGET_BUILD_DIR)/image-gateware+bios+none.bin
+
+gateware-flash-$(PLATFORM): $(GATEWARE_BIOS_FILE)
+	iceprog $(GATEWARE_BIOS_FILE)
+
+# To avoid duplicating the mkimage.py call here, if the user has not
```
mithro: Why isn't this needed for Mimas and others?

.travis.yml
cr1901: Because I'm on the fence about which version we should use; I _really_ want the cache of `lm32.lite`, but it makes the design more difficult to route. `lm32.minimal` is indeed easier to route, but lacks cache.

cr1901: Perhaps it might. I guess for now I'll disable `lm32.lite` from the build matrix, but I will be actively testing in locally.

PR #93, opened 2018-10-29
Add RGB LED support to ICE40UP5K-B-EVN Target

cr1901:
This PR updates submodules to pull in UART FIFO improvements, adds RGB LED support to the `ControlAndStatus` module, and then adds the newly updated `ControlAndStatus` module to the ICE40UP5K-B-EVN `base` target.

Tested with `lm32.minimal` CPU variant.

### DIFF
--- a/gateware/cas.py
+++ b/gateware/cas.py
@@ -10,6 +10,7 @@ from litex.soc.interconnect.csr_eventmanager import *
 from migen.genlib.misc import WaitTimer
 
 from litex.soc.cores.gpio import GPIOIn, GPIOOut
+from gateware.led import RGBLed
 
 class ControlAndStatus(Module, AutoCSR):
     def __init__(self, platform, clk_freq):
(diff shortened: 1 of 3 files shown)


gateware/cas.py
```diff
@@ -10,9 +10,10 @@
 from migen.genlib.misc import WaitTimer
 
 from litex.soc.cores.gpio import GPIOIn, GPIOOut
+from gateware.led import RGBLed
 
 class ControlAndStatus(Module, AutoCSR):
-    def __init__(self, platform, clk_freq):
+    def __init__(self, platform, clk_freq, rgb_pwm=False):
```
mithro: What is with you and random boolean flags given to modules? :-P

I think we probably want to consider if we want pwm for *all* LEDs rather than only *just* for rgb LEDs?

Do you think you could split the RGB led and PWM parts of this pull request apart?

platforms/ice40_up5k_b_evn.py
```diff
@@ -31,13 +31,13 @@
 
 # The ICE40UP5K-B-EVN does not use the provided FT2232H chip to provide a
 # UART port. One must use their own USB-to-serial cable instead to get a UART.
-# We have chosen to use 37A and 36B for "tx" and "rx" respectively on Header B
+# We have chosen to use 48B and 51A for "tx" and "rx" respectively on Header B
 # to implement UART connections. The board comes unpopulated and will need to
 # have headers soldered.
```
mithro: Why the change here?

platforms/ice40_up5k_b_evn.py
```diff
@@ -31,13 +31,13 @@
 
 # The ICE40UP5K-B-EVN does not use the provided FT2232H chip to provide a
```
mithro: Could we move the UART port to the "PMOD socket" and make sure things "just work" if you plug in a [Digilent PmodUSBUART](https://store.digilentinc.com/pmod-usbuart-usb-to-uart-interface/)?

I'm pretty sure I shipped you that Pmod, right?

platforms/ice40_up5k_b_evn.py
cr1901: Pins 37A and 37B are shared with switches on this board, so I have to choose either one functionality or the other.

PR #126, opened 2019-04-24
scripts: Add build-zephyr.sh script.

mateusz-holenko:
This adds a script for building Zephyr RTOS for VexRiscv as a firmware for LiteX Buildenv.

### DIFF
--- a/scripts/build-zephyr.sh
+++ b/scripts/build-zephyr.sh
@@ -1,0 +1,97 @@
+#!/bin/bash
+
+if [ "`whoami`" = "root" ]
+then
+	echo "Running the script as root is not permitted"
+	exit 1
+fi
+
+CALLED=$_
+[[ "${BASH_SOURCE[0]}" != "${0}" ]] && SOURCED=1 || SOURCED=0
+
+SCRIPT_SRC=$(realpath ${BASH_SOURCE[0]})
+SCRIPT_DIR=$(dirname $SCRIPT_SRC)
+TOP_DIR=$(realpath $SCRIPT_DIR/..)
+
+if [ $SOURCED = 1 ]; then
+	echo "You must run this script, rather then try to source it."
+	echo "$SCRIPT_SRC"
+	exit 1
+fi
+
+if [ -z "$HDMI2USB_ENV" ]; then
+	echo "You appear to not be inside the HDMI2USB environment."
+	echo "Please enter environment with:"
+	echo "  source scripts/enter-env.sh"
+	exit 1
+fi
+
+# Imports TARGET, PLATFORM, CPU and TARGET_BUILD_DIR from Makefile
+eval $(make env)
+make info
+
+source $SCRIPT_DIR/settings.sh
+
+# local version of SDK has a higher priority as it might contain a newer SDK
+SDK_LOCAL_LOCATION="$BUILD_DIR/zephyr_sdk"
+if [ -d "$SDK_LOCAL_LOCATION" ]; then
+	export ZEPHYR_SDK_INSTALL_DIR="$SDK_LOCAL_LOCATION"
+elif [ -z "$ZEPHYR_SDK_INSTALL_DIR" ]; then
+	echo "Zephyr SDK not found"
+	echo "Did you forget to run scripts/download-env.sh?"
+	exit 1
+fi
+
+export ZEPHYR_TOOLCHAIN_VARIANT=zephyr
+echo "Using Zephyr SDK from: $ZEPHYR_SDK_INSTALL_DIR"
+
+set -x
+set -e
+
+ZEPHYR_REPO=https://github.com/zephyrproject-rtos/zephyr
+ZEPHYR_REPO_BRANCH=master
+
+case $CPU in
+	vexriscv)
+		case "$CPU_VARIANT" in
+			lite* | standard* | full* | linux*)
+				TARGET_BOARD=arty_litex_vexriscv
+				ZEPHYR_REPO=https://github.com/antmicro/zephyr
+				ZEPHYR_REPO_BRANCH=litex-vexriscv
+				;;
+			*)
+				echo "Zephyr needs a CPU_VARIANT set to at least 'lite' for the support of 'ecall' instruction."
+				echo "Supported variants: lite, lite+debug, standard, standard+debug, full, full+debug, linux."
+				echo "Currently selected variant: $CPU_VARIANT"
+				exit 1
+				;;
+		esac
+		;;
+	*)
+		echo "CPU $CPU isn't supported at the moment."
+		exit 1
+		;;
+esac
+
+if [ "$FIRMWARE" != "zephyr" ]; then
+	echo "When building Zephyr you should set FIRMWARE to 'zephyr'."
+	exit 1
+fi
+
+ZEPHYR_SRC_DIR=$THIRD_DIR/zephyr
+ZEPHYR_APP=${ZEPHYR_APP:-subsys/shell/shell_module}
+OUTPUT_DIR=$TOP_DIR/$TARGET_BUILD_DIR/software/zephyr
+export ZEPHYR_BASE=$ZEPHYR_SRC_DIR/zephyr
+
+if [ ! -d "$ZEPHYR_SRC_DIR" ]; then
+	mkdir -p $ZEPHYR_SRC_DIR
+	cd $ZEPHYR_SRC_DIR
+	west init --manifest-url $ZEPHYR_REPO --manifest-rev $ZEPHYR_REPO_BRANCH
+fi
+west build -b $TARGET_BOARD $ZEPHYR_SRC_DIR/zephyr/samples/$ZEPHYR_APP --build-dir $OUTPUT_DIR
+
+cd $OUTPUT_DIR
+if [ ! -f "firmware.bin" ]; then
+	ln -s zephyr/zephyr.bin firmware.bin
+fi
+
--- a/scripts/download-env.sh
+++ b/scripts/download-env.sh
@@ -383,6 +383,36 @@ check_version ${CPU_ARCH}-elf-gcc $GCC_VERSION
 #conda install -y $CONDA_FLAGS gdb-${CPU_ARCH}-elf=$GDB_VERSION
 #check_version ${CPU_ARCH}-elf-gdb $GDB_VERSION
 
+# Zephyr SDK
+################################################
+if [ "$FIRMWARE" = "zephyr" ]; then
+	SCRIPT_NAME="zephyr-sdk-$ZEPHYR_SDK_VERSION-setup.run"
+	LOCAL_LOCATION="$BUILD_DIR/zephyr_sdk"
+
+	DETECTED_SDK_LOCATION=""
+	for DIR in "$LOCAL_LOCATION" "$ZEPHYR_SDK_INSTALL_DIR"; do
+		if [ -d "$DIR" ]; then
+			cat "$DIR/sdk_version" | grep -q $ZEPHYR_SDK_VERSION && DETECTED_SDK_LOCATION="$DIR"
+		fi
+	done
+
+	echo
+	if [ -d "$DETECTED_SDK_LOCATION" ]; then
+		echo "Zephyr SDK $ZEPHYR_SDK_VERSION found in: $DETECTED_SDK_LOCATION"
+		echo "---------------------------------------"
+	else
+		echo "Installing Zephyr SDK $ZEPHYR_SDK_VERSION"
+		echo "---------------------------------------"
+		(
+			mkdir -p "$LOCAL_LOCATION"
+			wget https://github.com/zephyrproject-rtos/sdk-ng/releases/download/v$ZEPHYR_SDK_VERSION/$SCRIPT_NAME -O "$LOCAL_LOCATION/$SCRIPT_NAME"
+			cd "$LOCAL_LOCATION"
+			chmod u+x $SCRIPT_NAME
+			./$SCRIPT_NAME -- -y -d "$PWD"
+			rm $SCRIPT_NAME
+		)
+	fi
+fi
 
 # Python modules
 ################################################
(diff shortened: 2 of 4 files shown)


scripts/build-zephyr.sh
```diff
@@ -0,0 +1,100 @@
...
+
+if [ -z "$HDMI2USB_ENV" ]; then
+	echo "You appear to not be inside the HDMI2USB environment."
+	echo "Please enter environment with:"
+	echo "  source scripts/enter-env.sh"
+	exit 1
+fi
+
+if [ -z "$ZEPHYR_SDK_INSTALL_DIR" ]; then
+	echo "Zephyr SDK not found"
```
mithro: Can you make this script download Zephyr SDK?
Also would it be better if this checked out a git repository like the Linux script does?
IE I think people running this script will probably want to hack on zephyr itself?

scripts/build-zephyr.sh
```diff
@@ -0,0 +1,100 @@
...
+
+set -x
+set -e
+
+ZEPHYR_REPO=https://github.com/zephyrproject-rtos/zephyr
+ZEPHYR_REPO_BRANCH=master
+
+case $CPU in
+	vexriscv)
+		TARGET_BOARD=arty_litex_vexriscv
```
mithro: This shouldn't be set? Zephyr should run on any board which supports litex right?

scripts/build-zephyr.sh
```diff
@@ -0,0 +1,100 @@
...
+		ZEPHYR_REPO=https://github.com/antmicro/zephyr
+		ZEPHYR_REPO_BRANCH=litex-vexriscv
+		;;
```
mithro: I think you might want to check `CPU_VARIANT`?

scripts/build-zephyr.sh
```diff
@@ -0,0 +1,100 @@
...
+
+cd $ZEPHYR_SRC_DIR
+
+export ZEPHYR_TOOLCHAIN_VARIANT=zephyr
+source zephyr-env.sh
+
+cd samples/$ZEPHYR_APP
+mkdir build || true
+cd build
+cmake -DBOARD=$TARGET_BOARD ..
```
mithro: It would be better if this output ended up under the `$OUTPUT_DIR` below... Maybe create a symlink?

scripts/build-zephyr.sh
futaris: litex-buildenv should make it a little easier for the end user, and maybe install the ZEPHYR_SDK.

```
$ wget https://github.com/zephyrproject-rtos/meta-zephyr-sdk/releases/download/0.10.0/zephyr-sdk-0.10.0-setup.run
$ chmod +x zephyr-sdk-0.10.0-setup.run
$ ./zephyr-sdk-0.10.0-setup.run
```

https://docs.zephyrproject.org/latest/getting_started/installation_linux.html
https://github.com/zephyrproject-rtos/meta-zephyr-sdk/

Not sure that grabbing the source & building the SDK is the best option.

It's built in a docker, using poky (openembedded)...  You can build without the SDK, but I haven't tried it.

mithro:
You'll need to rebase to make Travis work.

scripts/build-zephyr.sh
futaris: I got it running on minerva, in renode. Haven't tested on picorv32.

mithro:
I rebased this onto master to make travis go green.

It would be good if we could figure out how to make this run on travis.

scripts/build-zephyr.sh
mateusz-holenko: I think downloading and unpacking the SDK in `download-env.sh` if no SDK is detected in the system is the simplest option.

Using pre-built SDK lets you to do hacking in Zephyr anyway.

scripts/build-zephyr.sh
mateusz-holenko: PicoRV32 is a bit different as it has non-standard interrupt handling and you need a custom Zephyr port for it.

scripts/build-zephyr.sh
mithro: Can you download the SDK into the `./build/` directory?

scripts/build-zephyr.sh
mateusz-holenko: > This shouldn't be set? Zephyr should run on any board which supports litex right?

I understand you mean `arty` part, right?

You are right that it would be much nicer to have just `TARGET_BOARD=litex_vexriscv`, but `arty` dependency is needed as it provides the amount of RAM.

I think it is a good idea to rename the board so that `arty` part is not so prominent (to sth like `litex_vexriscv_arty`), but I'm not sure if we can get rid of it completely.

scripts/build-zephyr.sh
mateusz-holenko: I switched to `west` script that allows to build outside Zephyr tree, so no moves/symlinks are needed.

scripts/build-zephyr.sh
mateusz-holenko: There are many acceptable `CPU_VARIANTS`, so the validating `if` is quite long. 
I think, however, that it's better to explicitely list all acceptable options rather than fail on `minimal`/`minimal+debug` variants.

scripts/build-zephyr.sh
mateusz-holenko: The script downloads SDK to `build/zephyr_sdk`.

scripts/build-zephyr.sh
mithro: You can't hardcode the amount of memory based on the target. You need to get the amount of memory from the litex generated memory map. See [regions.ld](https://github.com/timvideos/HDMI2USB-firmware-prebuilt/blob/master/archive/master/v0.0.4-455-gde59df1/arty/base/vexriscv/software/include/generated/regions.ld) and [mem.h](https://github.com/timvideos/HDMI2USB-firmware-prebuilt/blob/master/archive/master/v0.0.4-455-gde59df1/arty/base/vexriscv/software/include/generated/mem.h).

If there needs to be a certain amount of RAM, we should be checking that.

mateusz-holenko: In general you are right, LiteX (even assuming that we simplify it to one type of core, VexRiscv in this case) is dynamic and highly configurable so there is no *one* LiteX+VexRiscv platform.

We need to provide *some* default configuration of LiteX, though, so it's possible to build Zephyr without building LiteX itself.
Recreating LiteX defaults seems to be a sane choice and this is where `arty_litex_vexriscv` came from.

What we are missing currently is the ability to **reconfigure** the platform dynamically.
This is something we have to address before merging the PR to Zephyr.
The options I currently see are:

* conditionally include `generated/mem.h` / `generated/csr.h` from LiteX in Zephyr's `board/riscv32/litex_vexriscv/board.h` and redefine there values read from the default `dts` (the one for LiteX+VexRiscv on Arty),

* conditionally include `generated/mem.h` / `generated/csr.h` from LiteX in the `dts` itself and use ifdefs and defines directly there,

* add a conditional include of custom `dtsi` file at the end of the default `dts` and pass a dynamic `dtsi` generated by `build-zephyr.sh` script (that would, of course, use data from headers generated by LiteX).

mateusz-holenko: We will proceed with this shortly, but in the meantime I believe it would be better to merge https://github.com/zephyrproject-rtos/zephyr/pull/14915 to master first as it has already been reviewed. Then we could issue another PR to Zephyr with a code making the LiteX platform more flexible. What do you think about it?

enjoy-digital: @mateusz-holenko: I'm not well aware of the efforts needed to get a PR merged in Zephyr so if it's easier for you, you could do this step by step and indeed address the genericity in another PR. We could try to help on that too.

scripts/build-zephyr.sh
mithro: Great!

scripts/build-zephyr.sh
```diff
@@ -0,0 +1,106 @@
...
+
+case $CPU in
+	vexriscv)
+		if [ "$CPU_VARIANT" != "lite" -a \
+				"$CPU_VARIANT" != "lite+debug" -a \
+				"$CPU_VARIANT" != "standard" -a \
+				"$CPU_VARIANT" != "standard+debug" -a \
+				"$CPU_VARIANT" != "full" -a \
+				"$CPU_VARIANT" != "full+debug" -a \
+				"$CPU_VARIANT" != "linux" ]; then
```
mithro: This is probably better as a case statement. https://linuxize.com/post/bash-case-statement/
```bash
case "$CPU_VARIANT" in
   "lite*" | "standard*" | "full*" | "linux*")
       ;;
  *)
	echo "Zephyr needs a CPU_VARIANT set to at least 'lite' for the support of 'ecall' instruction."
	echo "Supported variants: lite, lite+debug, standard, standard+debug, full, full+debug, linux."
	exit 1
	;;
esac
```

scripts/build-zephyr.sh
mithro: I think either of your last top options make sense.

BTW I started (but never finished) a tool to take a `csr.cvs` file and generate a [Linux Device Tree file](https://github.com/mithro/litex-devicetree) - see https://github.com/mithro/litex-devicetree

scripts/build-zephyr.sh
mithro: I think that makes sense.

scripts/build-zephyr.sh
```diff
@@ -0,0 +1,106 @@
...
+if [ "$FIRMWARE" != "zephyr" ]; then
+	echo "When building Zephyr you should set FIRMWARE to 'zephyr'."
+	exit 1
+fi
+
+function check_python_package {
+	python -c "import $1"
+}
+
+check_python_package "yaml"     || conda install pyyaml
```
mithro: These three python packages should maybe be part of the `download-env.sh` script too?

scripts/download-env.sh
```diff
@@ -383,6 +383,37 @@ check_version ${CPU_ARCH}-elf-gcc $GCC_VERSION
 #conda install -y $CONDA_FLAGS gdb-${CPU_ARCH}-elf=$GDB_VERSION
 #check_version ${CPU_ARCH}-elf-gdb $GDB_VERSION
 
+# Zephyr SDK
+################################################
+if [ "$FIRMWARE" = "zephyr" ]; then
+	VERSION="0.10.0"
```
mithro: This version should go in https://github.com/timvideos/litex-buildenv/blob/master/scripts/settings.sh

scripts/build-zephyr.sh
mateusz-holenko: That's a much nicer notation!

scripts/build-zephyr.sh
mateusz-holenko: moved

scripts/download-env.sh
mateusz-holenko: moved

mateusz-holenko:
I noticed problems with `west init` command (west is a toolbox for building Zephyr). 

It turns out that there is some configuration for `west` shipped in the repo together with Zephyr. Our Zephyr's branch is a bit behind master (as we are still waiting for merging https://github.com/zephyrproject-rtos/zephyr/pull/15418). The configuration there is for an older version of `west` and a result it fails. I checked and it works for master, so after rebasing our branch all will be back to normal.

In other words it's not the problem with this script, it's the problem with our branch that will be rebased anyway.

mateusz-holenko:
> It would be good if we could figure out how to make this run on travis.

The most basic test would be to configure the environment and build zephyr on some stock Linux image (so that we make sure all required dependencies, including Zephyr SDK, are actually downloaded and installed locally):
```
$ export CPU=vexriscv; export FIRMWARE=zephyr; export TARGET=net; export PLATFORM=arty
$ export CPU_VARIANT=lite
$ ./scripts/download-env.sh
$ source ./scripts/enter-env.sh
$ ./scripts/build-zephyr.sh
```

I don't know your current travis setup, but we could use`docker`/`singularity` for this.

scripts/download-env.sh
```diff
@@ -425,6 +455,24 @@ echo "Installing HDMI2USB-mode-switch (flashing and config tool)"
 pip install --upgrade git+https://github.com/timvideos/HDMI2USB-mode-switch.git
 check_import_version hdmi2usb.modeswitch $HDMI2USB_MODESWITCH_VERSION
 
+if [ "$FIRMWARE" = "zephyr" ]; then
```
mithro: These need to be in `enter-env.sh` script too.

You should be able to load `enter-env.sh` and `download-env.sh` in a side-by-side diff tool (like vimdiff) and see the comparisons.....

mithro:
> > It would be good if we could figure out how to make this run on travis.
> 
> The most basic test would be to configure the environment and build zephyr on some stock Linux image (so that we make sure all required dependencies, including Zephyr SDK, are actually downloaded and installed locally):
> 
> ```
> $ export CPU=vexriscv; export FIRMWARE=zephyr; export TARGET=net; export PLATFORM=arty
> $ export CPU_VARIANT=lite
> $ ./scripts/download-env.sh
> $ source ./scripts/enter-env.sh
> $ ./scripts/build-zephyr.sh
> ```
> 
> I don't know your current travis setup, but we could use`docker`/`singularity` for this.

Travis is configured in the `.travis.yaml` file at https://github.com/timvideos/litex-buildenv/blob/master/.travis.yml
The scripts it runs can be found under the `.travis` directory at https://github.com/timvideos/litex-buildenv/tree/master/.travis

Should be as simple as adding a new line or two to the .travis.yml file and then adding Zephyr here -> https://github.com/timvideos/litex-buildenv/blob/005f7463bd5e022ee3e3d8cc90efde57e8465b86/.travis/build.sh#L177-L204
and here -> 
https://github.com/timvideos/litex-buildenv/blob/005f7463bd5e022ee3e3d8cc90efde57e8465b86/.travis/build.sh#L251

BTW You can enable travis on your own fork via going to the travis-ci.org website.

PR #127, opened 2019-04-24
scripts: Add Renode emulation support.

mateusz-holenko:
This adds support for running LiteX in Renode emulation framework.

### DIFF
--- a/scripts/build-renode.sh
+++ b/scripts/build-renode.sh
@@ -1,0 +1,90 @@
+#!/bin/bash
+
+SCRIPT_SRC=$(realpath ${BASH_SOURCE[0]})
+SCRIPT_DIR=$(dirname $SCRIPT_SRC)
+TOP_DIR=$(realpath $SCRIPT_DIR/..)
+if [[ "${BASH_SOURCE[0]}" != "${0}" ]]; then
+	echo "You must run this script, rather then try to source it."
+	echo "$SCRIPT_SRC"
+	exit 1
+fi
+
+source $SCRIPT_DIR/build-common.sh
+
+init
+
+RENODE_BIN=${RENODE_BIN:-renode}
+if ! command -v $RENODE_BIN 2>&1 1>/dev/null; then
+	# Download prebuilt renode Release if none is currently installed
+
+	RENODE_PACKAGE=renode-latest.pkg.tar.xz
+	RENODE_URL=https://antmicro.com/projects/renode/builds/$RENODE_PACKAGE
+	RENODE_LOCATION="$BUILD_DIR/renode"
+	RENODE_BIN=$RENODE_LOCATION/opt/renode/bin/Renode.exe
+
+	if [ ! -x $RENODE_BIN ]; then
+		mkdir -p $RENODE_LOCATION
+		(
+			cd $RENODE_LOCATION
+			wget $RENODE_URL
+			tar -xf $RENODE_PACKAGE
+		)
+		
+		chmod u+x $RENODE_BIN
+		echo "Renode downloaded and installed locally: $RENODE_BIN" 
+	fi
+fi
+
+case $CPU in
+	vexriscv | picorv32)
+		;;
+	*)
+		echo "CPU $CPU_TYPE isn't supported at the moment."
+		exit 1
+		;;
+esac
+
+LITEX_CONFIG_FILE="$TARGET_BUILD_DIR/test/csr.csv"
+if [ ! -f "$LITEX_CONFIG_FILE" ]; then
+	make firmware
+fi
+
+# Ethernet
+ETH_BASE_ADDRESS=$(parse_generated_header "csr.h" CSR_ETHMAC_BASE)
+if [ ! -z "$ETH_BASE_ADDRESS" ]; then
+	RENODE_NETWORK=${RENODE_NETWORK:-tap}
+	case $RENODE_NETWORK in
+	tap)
+		echo "Using tun device for Renode networking, (may need sudo)..."
+		configure_tap
+		start_tftp
+
+		# Build/copy the image into the TFTP directory.
+		make tftp
+
+		TAP_INTERFACE=tap0
+		;;
+
+	none)
+		echo "Renode networking disabled..."
+		;;
+	*)
+		echo "Unknown RENODE_NETWORK mode '$RENODE_NETWORK'"
+		return 1
+		;;
+	esac
+fi
+
+RENODE_SCRIPTS_DIR="$TARGET_BUILD_DIR/renode"
+RENODE_RESC="$RENODE_SCRIPTS_DIR/litex_buildenv.resc"
+RENODE_REPL="$RENODE_SCRIPTS_DIR/litex_buildenv.repl"
+
+mkdir -p $RENODE_SCRIPTS_DIR
+python $SCRIPT_DIR/generate-renode-scripts.py $LITEX_CONFIG_FILE \
+	--repl "$RENODE_REPL" \
+	--resc "$RENODE_RESC" \
+	--bios-binary "$TARGET_BUILD_DIR/software/bios/bios.bin" \
+	--configure-network ${TAP_INTERFACE:-""}
+
+$RENODE_BIN "$RENODE_RESC"
+
--- a/scripts/generate-renode-scripts.py
+++ b/scripts/generate-renode-scripts.py
@@ -1,0 +1,287 @@
+#!/usr/bin/env python3
+"""
+Copyright (c) 2019 Antmicro
+
+Renode platform definition (repl) and script (resc) generator for LiteX SoC.
+
+This script parses LiteX 'csr.csv' file and generates scripts for Renode
+necessary to emulate the given configuration of the LiteX SoC.
+"""
+
+import sys
+import csv
+import argparse
+
+# this memory region is defined and handled
+# directly by LiteEth model in Renode
+non_generated_mem_regions = ['ethmac']
+
+mem_regions = {}
+peripherals = {}
+constants = {}
+
+def generate_ethmac(peripheral, **kwargs):
+    """ Generates definition of 'ethmac' peripheral.
+
+    Args:
+        peripheral (dict): peripheral description
+        kwargs (dict): additional parameters, including 'buffer'
+
+    Returns:
+        string: repl definition of the peripheral
+    """
+    buf = kwargs['buffer']()
+
+    result = """
+ethmac: Network.LiteX_Ethernet @ {{
+    sysbus <{}, +0x100>;
+    sysbus new Bus.BusMultiRegistration {{ address: {}; size: {}; region: "buffer" }}
+}}
+""".format(peripheral['address'], buf['address'], buf['size'])
+
+    if 'interrupt' in peripheral['constants']:
+        result += '    -> cpu@{}\n'.format(peripheral['constants']['interrupt'])
+
+    return result
+
+def generate_memory_region(region_descriptor):
+    """ Generates definition of memory region.
+
+    Args:
+        region_descriptor (dict): memory region description
+
+    Returns:
+        string: repl definition of the memory region
+    """
+
+    return """
+{}: Memory.MappedMemory @ sysbus {}
+    size: {}
+""".format(region_descriptor['name'], region_descriptor['address'], region_descriptor['size'])
+
+def generate_silencer(peripheral, **kwargs):
+    """ Silences access to a memory region.
+
+    Args:
+        peripheral (dict): peripheral description
+        kwargs (dict): additional parameters, not used
+
+    Returns:
+        string: repl definition of the silencer
+    """
+    return """
+sysbus:
+    init add:
+        SilenceRange <{} 0x200> # {}
+""".format(peripheral['address'], peripheral['name'])
+
+def generate_cpu():
+    """ Generates definition of a CPU.
+
+    Returns:
+        string: repl definition of the CPU
+    """
+    kind = constants['config_cpu_type']['value']
+
+    if kind == 'VEXRISCV':
+        return """
+cpu: CPU.VexRiscv @ sysbus
+"""
+    elif kind == 'PICORV32':
+        return """
+cpu: CPU.PicoRV32 @ sysbus
+    cpuType: "rv32imc"
+"""
+    else:
+        raise Exception('Unsupported cpu type: {}'.format(kind))
+
+def generate_peripheral(peripheral, **kwargs):
+    """ Generates definition of a peripheral.
+
+    Args:
+        peripheral (dict): peripheral description
+        kwargs (dict): additional parameters, including 'model' and 'properties'
+
+    Returns:
+        string: repl definition of the peripheral
+    """
+    result = '\n{}: {} @ sysbus {}\n'.format(
+        peripheral['name'],
+        kwargs['model'],
+        peripheral['address'])
+
+    for constant, val in peripheral['constants'].items():
+        if constant == 'interrupt':
+            result += '    -> cpu@{}\n'.format(val)
+        else:
+            result += '    {}: {}\n'.format(constant, val)
+
+    if 'properties' in kwargs:
+        for prop, val in kwargs['properties'].items():
+            result += '    {}: {}\n'.format(prop, val())
+
+    return result
+
+def generate_repl():
+    """ Generates platform definition.
+
+    Returns:
+        string: platform defition containing all supported peripherals and memory regions
+    """
+    result = ""
+
+    # defines mapping of LiteX peripherals to Renode models
+    name_to_handler = {
+        'uart': {
+            'handler': generate_peripheral,
+            'model': 'UART.LiteX_UART'
+        },
+        'timer0': {
+            'handler': generate_peripheral,
+            'model': 'Timers.LiteX_Timer',
+            'properties': {
+                'frequency': lambda: constants['system_clock_frequency']['value']
+            }
+        },
+        'ethmac': {
+            'handler': generate_ethmac,
+            'buffer': lambda: mem_regions['ethmac']
+        },
+        'ddrphy': {
+            'handler': generate_silencer
+        },
+        'sdram': {
+            'handler': generate_silencer
+        },
+        'ethphy': {
+            'handler': generate_silencer
+        }
+    }
+
+    for mem_region in mem_regions.values():
+        if mem_region['name'] not in non_generated_mem_regions:
+            result += generate_memory_region(mem_region)
+
+    result += generate_cpu()
+
+    for name, peripheral in peripherals.items():
+        if name not in name_to_handler:
+            print('Skipping unsupported peripheral {} at {}'.format(name, peripheral['address']))
+            continue
+
+        h = name_to_handler[name]
+        result += h['handler'](peripheral, **h)
+
+    return result
+
+def parse_csv(data):
+    """ Parses LiteX CSV file.
+
+    Args:
+        data (list): list of CSV file lines
+    """
+
+    # scan for CSRs first, so it's easier to resolve CSR-related constants in the second pass
+    for _type, _name, _address, _, __ in data:
+        if _type == 'csr_base':
+            peripherals[_name] = {'name': _name, 'address': _address, 'constants': {}}
+
+    for _type, _name, _val, _val2, _ in data:
+        if _type == 'csr_base':
+            # CSRs have already been parsed
+            pass
+        elif _type == 'csr_register':
+            # we are currently not interested in this
+            pass
+        elif _type == 'constant':
+            found = False
+            for _csr_name in peripherals:
+                if _name.startswith(_csr_name):
+                    peripherals[_csr_name]['constants'][_name[len(_csr_name)+1:]] = _val
+                    found = True
+                    break
+            if not found:
+                # if it's not a CSR-related constant, it must be a global one
+                constants[_name] = {'name': _name, 'value': _val}
+        elif _type == 'memory_region':
+            mem_regions[_name] = {'name': _name, 'address': _val, 'size': _val2}
+        else:
+            print('Skipping unexpected CSV entry: {} {}'.format(_type, _name))
+
+def generate_resc(repl_file, host_tap_interface=None, bios_binary=None):
+    """ Generates platform definition.
+
+    Args:
+        repl_file (string): path to Renode platform definition file
+        host_tap_interface (string): name of the tap interface on host machine or None if no network should be configured
+        bios_binary (string): path to the binary file of LiteX BIOS or None if it should not be loaded into ROM
+
+    Returns:
+        string: platform defition containing all supported peripherals and memory regions
+    """
+    cpu_type = constants['config_cpu_type']['value']
+
+    result = """
+using sysbus
+mach create "litex-{}"
+machine LoadPlatformDescription @{}
+cpu StartGdbServer 10001
+showAnalyzer sysbus.uart
+showAnalyzer sysbus.uart Antmicro.Renode.Analyzers.LoggingUartAnalyzer
+""".format(cpu_type, repl_file)
+
+    rom_base = mem_regions['rom']['address']
+    if rom_base is not None and bios_binary:
+        # load LiteX BIOS to ROM
+        result += """
+sysbus LoadBinary @{} {}
+cpu PC {}
+""".format(bios_binary, rom_base, rom_base)
+
+    if host_tap_interface:
+        # configure network to allow netboot
+        result += """
+emulation CreateSwitch "switch"
+emulation CreateTap "{}" "tap"
+connector Connect ethmac switch
+connector Connect host.tap switch
+""".format(host_tap_interface)
+
+    result += 'start'
+    return result
+
+def print_or_save(filepath, lines):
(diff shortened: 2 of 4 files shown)


mithro:
The conda issue should be fixed by https://github.com/timvideos/litex-buildenv/pull/129

futaris:
Not sure that litex_picorv32 external interrupts are hooked up correctly, particularly for TARGET=base

```
export CPU=picorv32 PLATFORM=arty TARGET=net
source ./scripts/enter-env.sh
./scripts/build/renode.sh
```
works

```
export CPU=picorv32 PLATFORM=arty TARGET=base
source ./scripts/enter-env.sh
./scripts/build/renode.sh
```
hangs

It also hangs if you remove the eth node from /opt/renode/platforms/cpus/litex_picorv32.repl ...

mithro:
You'll need to rebase to make Travis work.

mateusz-holenko:
> Not sure that litex_picorv32 external interrupts are hooked up correctly, particularly for TARGET=base
> 
> ```
> export CPU=picorv32 PLATFORM=arty TARGET=net
> source ./scripts/enter-env.sh
> ./scripts/build/renode.sh
> ```
> works

It works only partially as you can't write on UART.

> 
> ```
> export CPU=picorv32 PLATFORM=arty TARGET=base
> source ./scripts/enter-env.sh
> ./scripts/build/renode.sh
> ```
> hangs
> 
> It also hangs if you remove the eth node from /opt/renode/platforms/cpus/litex_picorv32.repl ...

I verified the irq mapping in `platforms/cpus/litex_picorv32.repl` and you are right, it's not compatible with the defaults generated by LiteX, but...

I believe that there is a problem with overlapping interrupts in LiteX + PicoRV32.
LiteX uses IRQ1 for timer, but in PicoRV32 external IRQ1 is shared with internal `ecall/ebreak/illegal instruction`.
This caused problems when I was working on Zephyr, so I modified LiteX to use different set of IRQs for PicoRV32.
The problem was I hadn't created any PR with the change in LiteX, but published the non-standard mapping in Renode.

To sort it out, I've just crated a PR in [LiteX](https://github.com/cliffordwolf/picorv32#custom-instructions-for-irq-handling).

There is an issues though, BuildEnv provides its own mapping of peripheral's IRQs and for modified LiteX there is a conflict in `targets/arty/net.py` (and possibly other files as well).
I fixed it locally by changing `ethmac` IRQ to 6 (and this is the value in Renode's `repl` file), but I expect more fixes are needed to make it right.

mithro:
I rebased this onto master to get travis to run.

futaris:
https://github.com/enjoy-digital/litex/pull/174 for the issue mentioned above.

scripts/build-renode.sh
```diff
@@ -0,0 +1,170 @@
...
+}
+
+add_to_script "using sysbus"
+add_to_script "mach create \"litex-$CPU\""
+add_to_script "machine LoadPlatformDescription @platforms/cpus/$RENODE_PLATFORM"
+add_to_script "cpu StartGdbServer 10001"
+add_to_script "showAnalyzer sysbus.uart"
+
+# BIOS
+if grep -q 'ROM_BASE 0x00000000' $TARGET_BUILD_DIR/software/include/generated/mem.h; then
```
mithro: The ROM_BASE isn't always at memory address zero.

scripts/build-renode.sh
```diff
@@ -0,0 +1,170 @@
...
+	echo "Platform has BIOS ROM, adding BIOS"
+
+	add_to_script "sysbus LoadBinary @$TARGET_BUILD_DIR/software/bios/bios.bin 0x0"
+	add_to_script "cpu PC 0x0"
```
mithro: The reset vector isn't always zero.

scripts/build-renode.sh
mithro: https://github.com/timvideos/HDMI2USB-firmware-prebuilt/blob/52959d0bfe90b7536da7e516120c586e6ea3aafc/archive/master/v0.0.4-455-gde59df1/ice40_up5k_b_evn/base/lm32.minimal/software/include/generated/mem.h#L7

scripts/build-renode.sh
mithro: https://github.com/timvideos/HDMI2USB-firmware-prebuilt/blob/52959d0bfe90b7536da7e516120c586e6ea3aafc/archive/master/v0.0.4-455-gde59df1/ice40_up5k_b_evn/base/lm32.minimal/software/include/generated/csr.h#L322-L325

scripts/build-renode.sh
mateusz-holenko: You are right, I fixed it and now the actual value is read from `generated/mem.h`.

I took this from `scripts/build-qemu.sh` though, where this value is hard-coded.

scripts/build-renode.sh
mateusz-holenko: In the fixed version, I assumed the reset vector value to be the same as `ROM_BASE`.

pgielda:
I did some tests.
I would suggest adding two things:
1) Silencing SDRAM_CONTROLLER reads/writes so that the console is not flooded.
2) add "showAnalyzer uart Renode.Analyzers.LoggingUartAnalyzer" so that console output is printed with the rest of the log

pgielda:
We should also tag this peripheral:

02:17:37.1529 [WARNING] sysbus: [cpu: 0xC50] WriteDoubleWord to non existing peripheral at 0xE0005804, value 0x1.
02:17:37.1529 [WARNING] sysbus: [cpu: 0xC54] WriteDoubleWord to non existing peripheral at 0xE0005808, value 0x1.
02:17:37.1530 [WARNING] sysbus: [cpu: 0xC58] WriteDoubleWord to non existing peripheral at 0xE0005810, value 0x1.
02:17:37.1531 [WARNING] sysbus: [cpu: 0xC60] WriteDoubleWord to non existing peripheral at 0xE0005804, value 0x2.
02:17:37.1532 [WARNING] sysbus: [cpu: 0xC64] WriteDoubleWord to non existing peripheral at 0xE0005808, value 0x1.
02:17:37.1532 [WARNING] sysbus: [cpu: 0xC6C] WriteDoubleWord to non existing peripheral at 0xE0005810, value 0x1.

And most probably silence this too.

mateusz-holenko:
In the newest version:

- if Renode is not found in the system, version 1.7 is downloaded from github and copied to `build/renode` directory

- platform definition file is generated dynamically from `csr.h` and `mem.h` files

- access to some peripherals is silenced (so it does not flood the console)

- uart output is logged

scripts/build-renode.sh
```diff
@@ -0,0 +1,265 @@
...
+			ar x $RENODE_PACKAGE
+			tar -xf data.tar.gz
+		)
+		
+		chmod u+x $RENODE_BIN
+		echo "Renode downloaded and installed locally: $RENODE_BIN" 
+	fi
+fi
+
+script_file=$(mktemp /tmp/litex_buildenv.resc.XXXXXX)
```
mithro: Can we move the code generating the configuration to a Python script? It would be much nicer than this then?

scripts/build-renode.sh
```diff
@@ -0,0 +1,265 @@
...
+make info
+
+source $SCRIPT_DIR/settings.sh
+
+set -x
+set -e
+
+RENODE_BIN="renode"
+command -v $RENODE_BIN 2>&1 1>/dev/null
+if [ $? -ne 0 ]; then
```
mithro: Please add a comment here. I'm assuming this is downloading a prebuilt renode release? On what platforms will this run?

scripts/build-renode.sh
```diff
@@ -0,0 +1,265 @@
...
+# Imports TARGET, PLATFORM, CPU and TARGET_BUILD_DIR from Makefile
+eval $(make env)
+make info
+
+source $SCRIPT_DIR/settings.sh
+
+set -x
+set -e
+
+RENODE_BIN="renode"
```
mithro: Allow people to override the location of renode binary to use.

scripts/build-renode.sh
mateusz-holenko: You are right, this is a pre-built Renode compiled for 64-bit Linux. It should run on any distro assuming dependencies are present in the system (list of runtime dependecies is described [here](https://github.com/renode/renode/blob/master/README.rst#installation).

Other option is to install package (that would take care of dependencies automatically). The downside is that:

- it's installed globally in the system, polluting it a little bit,
- packages are distro-specific and we currently support: debian/ubuntu, fedora and arch so we would have to detect the distribution and install appropriate package; the question what about other distributions?

scripts/build-renode.sh
```diff
@@ -0,0 +1,162 @@
...
+		exit 1
+		;;
+esac
+
+LITEX_CONFIG_FILE="$TARGET_BUILD_DIR/test/csr.csv"
+if [ ! -f $LITEX_CONFIG_FILE ]; then
+	make firmware
+fi
+
+function parse_generated_header {
```
mithro: This is trying to get the IP address right? I would suggest we use the following name...
```suggestion
function get_ethernet_ip_address {
```

This is probably useful in the qemu emulation too. We should do that there too?

scripts/build-renode.sh
```diff
@@ -0,0 +1,162 @@
...
+		;;
+	*)
+		echo "Unknown RENODE_NETWORK mode '$RENODE_NETWORK'"
+		return 1
+		;;
+	esac
+fi
+
+script_file=$(mktemp /tmp/litex_buildenv.resc.XXXXXX)
+platform_file=$(mktemp /tmp/litex_buildenv.repl.XXXXXX)
```
mithro: Lets put these files in the build directory under `build/<build directory>/renode`?

mithro: We then don't need to clean them up and people can launch renode manually with the configurations.

scripts/generate-renode-scripts.py
```diff
@@ -0,0 +1,168 @@
+#!/usr/bin/env python3
+
```
mithro: Please add top level doc string.

scripts/generate-renode-scripts.py
```diff
@@ -0,0 +1,168 @@
...
+import csv
+import zlib
+import argparse
+
+def generate_ethmac(data):
```
mithro: Please add docstring and [maybe type hints](https://docs.python.org/3/library/typing.html).

scripts/generate-renode-scripts.py
```diff
@@ -0,0 +1,168 @@
...
+    base = get_entry_value(data, 'csr_base', 'ethmac')
+    buffer_base = get_entry_value(data, 'memory_region', 'ethmac')
+    buffer_size = get_entry_value(data, 'memory_region', 'ethmac', 1)
+    interrupt = get_entry_value(data, 'constant', 'ethmac_interrupt')
+
+    result = [
```
mithro: Use a triple quoted string for long output like this.

scripts/generate-renode-scripts.py
```diff
@@ -0,0 +1,168 @@
...
+        '    init add:',
+        '        SilenceRange <{} 0x200> # {}'.format(base, name)
+    ]
+
+def generate_flash_boot(data):
+    # TODO: this value is hardcoded, as I couldn't find it in the csv
+    base = 0x20228000
+    return [
+        'boot_flash: Memory.ArrayMemory @ sysbus {}'.format(base),
+        '    size: 0x00100000'
```
mithro: The size / base comes from `mem.h` in `build/XXXX/software/include/generated/`
```c
#define SPIFLASH_BASE 0xa0000000
#define SPIFLASH_SIZE 0x01000000
```

scripts/generate-renode-scripts.py
```diff
@@ -0,0 +1,168 @@
...
+
+def get_entry_value(data, kind, name, idx=0):
+    for entry in data:
+        if entry[0] == kind and entry[1] == name:
+            return entry[idx + 2]
+    return None
+
+def generate_repl(data):
+    return \
+        generate_cpu(data) + \
```
mithro: Don't use line continuations like this;
```python

def generate_repl(data):
 output = []
 output += generate_cpu(data)
 output += xxx
 return "".join(output)
```

scripts/generate-renode-scripts.py
```diff
@@ -0,0 +1,168 @@
...
+        '       sysbus <{}, +0x100>;'.format(base),
+        '       sysbus new Bus.BusMultiRegistration {{ address: {}; size: {}; region: "buffer" }}'.format(buffer_base, buffer_size),
+        '    }'
+    ]
+
+    if interrupt is not None:
+        result.append('    -> cpu@{}'.format(interrupt))
+    return result
+
+def generate_peripheral(data, name, model, properties=None):
```
mithro: What if the peripheral doesn't exist?

scripts/generate-renode-scripts.py
mithro: This is actually in the CSV file in `build/XXX/test/csr.csv` too;
```csv
memory_region,rom,0x00000000,32768,
memory_region,sram,0x10000000,32768,
memory_region,spiflash,0xa0000000,16777216,
memory_region,main_ram,0x40000000,268435456,
memory_region,ethmac,0xb0000000,8192,
```

scripts/generate-renode-scripts.py
```diff
@@ -0,0 +1,168 @@
...
+    for entry in data:
+        if entry[0] == kind and entry[1] == name:
+            return entry[idx + 2]
+    return None
+
+def generate_repl(data):
+    return \
+        generate_cpu(data) + \
+        generate_memory_region(data, 'rom') + \
+        generate_memory_region(data, 'sram') + \
```
mithro: You should loop over the memory regions define in either `mem.h` or `csr.csv`;
```python

data = """
memory_region,rom,0x00000000,32768,
memory_region,sram,0x10000000,32768,
memory_region,spiflash,0xa0000000,16777216,
memory_region,main_ram,0x40000000,268435456,
memory_region,ethmac,0xb0000000,8192,
"""
for line in csv.reader(data):
  if line[0] != "memory_region":
    _, memory_name, memory_start_adress, memory_size = *line
   # Some peripherals (like Ethernet) define a memory region, as peripherals will already be defined and taken ownership of the region we skip creating it here.
   if model.already_has_memory(memory_name):
      print("Skipping {} as already claimed by peripheral {}".format(memory_name, model.already_has_memory(memory_name))
      continue
    generate_memory_region(memory_name, memory_start_address, memory_size)
```

scripts/generate-renode-scripts.py
```diff
@@ -0,0 +1,168 @@
...
+        flash_boot_address = 0x20228000
+
+        firmware_data = open(firmware_binary, 'rb').read()
+        crc32 = zlib.crc32(firmware_data)
+
+        result.extend([
+            'sysbus WriteDoubleWord {} {}'.format(hex(flash_boot_address), hex(len(firmware_data))),
+            'sysbus WriteDoubleWord {} {}'.format(hex(flash_boot_address + 4), hex(crc32)),
+            'sysbus LoadBinary @{} {}'.format(firmware_binary, hex(flash_boot_address + 8))
+        ])
```
mithro: Can you add a comment about what this is doing and why?

scripts/generate-renode-scripts.py
```diff
@@ -0,0 +1,168 @@
...
+        raise Exception('Unsupported cpu type: {}'.format(kind))
+
+def get_entry_value(data, kind, name, idx=0):
+    for entry in data:
+        if entry[0] == kind and entry[1] == name:
+            return entry[idx + 2]
+    return None
+
+def generate_repl(data):
+    return \
```
mithro: You want code that does something like this;

```python
#! /usr/bin/env python3

data = """
csr_base,cas,0xe0006800,,
csr_base,ctrl,0xe0000000,,
csr_base,ddrphy,0xe0005800,,
csr_base,ethmac,0xe0007800,,
csr_base,ethphy,0xe0007000,,
csr_base,info,0xe0006000,,
csr_base,sdram,0xe0004000,,
csr_base,spiflash,0xe0005000,,
csr_base,timer0,0xe0002800,,
csr_base,uart,0xe0001800,,
csr_base,uart_phy,0xe0001000,,
csr_base,identifier_mem,0xe0002000,,
csr_register,cas_leds_out,0xe0006800,1,rw
csr_register,cas_switches_in,0xe0006804,1,ro
csr_register,cas_buttons_ev_status,0xe0006808,1,rw
csr_register,cas_buttons_ev_pending,0xe000680c,1,rw
csr_register,cas_buttons_ev_enable,0xe0006810,1,rw
constant,remoteip1,192,,
constant,remoteip2,168,,
constant,remoteip3,100,,
constant,remoteip4,100,,
constant,cas_leds_count,4,,
constant,cas_switches_count,4,,
constant,cas_buttons_count,4,,
constant,ethmac_rx_slots,2,,
constant,ethmac_tx_slots,2,,
"""

# Build a list of peripherals and collect all CSR registers + CSR constants associated with the peripherals
global_constants = {}
peripherals = {}
for line in csv.reader(data):
    if line[0] == "csr_base":
        _, peripheral_name, peripheral_register_base = line
        assert peripheral_name not in peripherals
        registers = {}
        constants = {}
        peripherals[peripheral_name] = (peripheral_register_base, registers, constants)

    elif line[0] == "csr_register":
        _, register_name, register_address, register_width, register_mode = line
        for peripheral_name in peripherals.keys():
            if not register_name.startswith(peripheral_name):
                continue
            peripheral_register_base, peripheral_registers, _ = peripherals[peripheral_name]

            # Strip the peripheral prefix
            register_short_name = constant_name[len(peripheral_name)+1:]

            register_offset = register_address - peripheral_register_base
            assert register_short_name not in peripheral_registers
            peripheral_registers[register_name] = (register_offset, register_width, register_mode)
            break
        else:
            # CSR Registers must be associated with a peripheral
            assert False, "CSR Register {} not associated with any peripheral {}".format(register_name, peripherals.keys())

    elif line[0] == "constant":
        _, constant_name, constant_value = line
        for peripheral_name in peripherals.keys():
            if not constant_name.startswith(peripheral_name):
                continue

            # Strip the peripheral prefix
            constant_short_name = constant_name[len(peripheral_name)+1:]

            peripheral_register_base, _, peripheral_constants = peripherals[peripheral_name]
            assert constant_short_name not in peripheral_constants
            peripheral_constants[constant_short_name] = constant_value

            break
        else:
            global_constants[constant_name] = constant_value

    else:
        assert False, "Unknown line '{}'".format(line)

import pprint
pprint.pprint(global_constants)
pprint.pprint(peripherals)

for peripheral_name, (peripheral_register_base, peripheral_registers, peripheral_constants) in peripherals.values():
    peripheral_handler = detect_peripheral_handler(peripheral_registers)
    if peripheral_handler is None:
        peripheral_hanlder = mock_handler(peripheral_register_base, peripheral_registers)

    peripheral_handler(config_file, peripheral_name, peripheral_register_base, peripheral_registers, peripheral_constants)
```

scripts/build-renode.sh
mateusz-holenko: I also moved generation of resc commands to the same python script.

mateusz-holenko:
> > Not sure that litex_picorv32 external interrupts are hooked up correctly, particularly for TARGET=base
> > ```
> > export CPU=picorv32 PLATFORM=arty TARGET=net
> > source ./scripts/enter-env.sh
> > ./scripts/build/renode.sh
> > ```
> > 
> > 
> > works
> 
> It works only partially as you can't write on UART.
> 
> > ```
> > export CPU=picorv32 PLATFORM=arty TARGET=base
> > source ./scripts/enter-env.sh
> > ./scripts/build/renode.sh
> > ```
> > 
> > 
> > hangs
> > It also hangs if you remove the eth node from /opt/renode/platforms/cpus/litex_picorv32.repl ...
> 
> I verified the irq mapping in `platforms/cpus/litex_picorv32.repl` and you are right, it's not compatible with the defaults generated by LiteX, but...
> 
> I believe that there is a problem with overlapping interrupts in LiteX + PicoRV32.
> LiteX uses IRQ1 for timer, but in PicoRV32 external IRQ1 is shared with internal `ecall/ebreak/illegal instruction`.
> This caused problems when I was working on Zephyr, so I modified LiteX to use different set of IRQs for PicoRV32.
> The problem was I hadn't created any PR with the change in LiteX, but published the non-standard mapping in Renode.
> 
> To sort it out, I've just crated a PR in [LiteX](https://github.com/cliffordwolf/picorv32#custom-instructions-for-irq-handling).
> 
> There is an issue though, BuildEnv provides its own mapping of peripheral's IRQs and for modified LiteX there is a conflict in `targets/arty/net.py` (and possibly other files as well).
> I fixed it locally by changing `ethmac` IRQ to 6 (and this is the value in Renode's `repl` file), but I expect more fixes are needed to make it right.

The ideas from my PR has been merged to LiteX master with [this commit](https://github.com/enjoy-digital/litex/commit/47dc87584f2bc29e837b4319f516ad3369473fd9).

@mithro I see that buildenv LIteX submodule is a bit behind master. Do you knwo if switching to a newer version would be a problem?

mithro:
@mateusz-holenko To update the submodules it should be as simple as going into `./third_party` directory and running `/merge-upstream.sh`.

scripts/build-renode.sh
mateusz-holenko: Before moving to a python script this function was used to extract different values defined in `mem.h`/`csr.h` files. Right now it's used in just one context, so it lost its universality - refactoring and renaming it seems like a good idea.

The whole tap-related fragment of script is a copy-paste from `build-qemu.sh`, so the two scripts shares a lot more code. I will extract it to some base file and source in both `build-renode.sh` and `build-qemu.sh`.

scripts/build-renode.sh
mateusz-holenko: This is a good idea.

scripts/generate-renode-scripts.py
mateusz-holenko: I was confused by the fact that in `include/generated/mem.h` `FLASH_BOOT_ADDRESS` is defined as `0x20228000`, but `spiflash` is located at `0xa0000000`.

Now I see in `targets/arty/base.py` that this address is a default shadow for `0x20000000`.
For Arty platform, the flash boot address is `spi_flash` base + gateware size (0x220000) + bios size (0x8000), so it all adds up.

It's still a bit problematic, as needed parameters (information about shadow address, gateware size or bios size) are not included in `csr.csv` so I don't see a way of calculating the final value based solely on this input.

scripts/generate-renode-scripts.py
mateusz-holenko: I created a separate parse_csv function that does the job. I skipped parsing registers as it's not needed for Renode.

scripts/generate-renode-scripts.py
mateusz-holenko: This fragment of code is temporarily removed. See comments above.

scripts/generate-renode-scripts.py
mateusz-holenko: I created a separate `parse_csv` function that does the job.

scripts/generate-renode-scripts.py
mateusz-holenko: My idea was to return an empty array that would concatenate nicely, but now the generating code is totally reworked, so this is no more valid.

scripts/generate-renode-scripts.py
mateusz-holenko: This fragment is totally reworked.

scripts/generate-renode-scripts.py
mateusz-holenko: I looked a little in LiteX sources and it looks that the best option is to emit more information (i.e., `shadow_base` and `flash_boot_offset`) to `csr.csv` file.

As it will require a separate PR to LiteX that would block this one, I decided to temporarily remove the functionality of booting from flash (still leaving a netboot option available). I will add it with a separate PR, once the necessary code is in LiteX.

scripts/generate-renode-scripts.py
mateusz-holenko: This is reworked to return a single, multiline string.

scripts/generate-renode-scripts.py
mateusz-holenko: I added docstrings to all functions.

scripts/generate-renode-scripts.py
mateusz-holenko: Done.

scripts/build-renode.sh
mateusz-holenko: It's done.

scripts/build-renode.sh
mateusz-holenko: After extracting the common function to `build-common.sh` it turns out that `parse_generated_header` function might be used in `build-qemu` in several places (in different context), so I kept it's name.

mateusz-holenko:
@mithro Did you have time to look at the current version of the script?

mateusz-holenko:
I rebased the code onto the current master and changed Renode binary source to nightly builds.

mateusz-holenko:
@mithro Travis failed, but it does not seem to be related to my changes. How can I force it to run once again?

PR #132, opened 2019-05-04
scripts: Rework bootstrap.

mithro:
* Check for the tools needed up `download-env.sh`.
 * Rename `download-env-root.sh` to `debian-setup.sh` (should stop people thinking we are Debian only).


mithro:
Woot! It is green now.

Issue #134, opened 2019-05-11
Change default board to iCE40 based board


ewenmcneill:
Now that we have a build chain that is fully open source (iCE40), we should change the default build target to a board that is iCE40 based.  Maybe TinyFPGA BX now, and iCEBreaker once those are out/supported?

The reasoning here is that getting started with a Xilinx based board requires several non trivial manual steps (installing Vivado, or ISE), which are completely unnecessary for the iCE40 based boards, but users that just start following the defaults can quickly end up in the weeds doing a lot of yak shaving (eg, https://twitter.com/femtoduino/status/1126750242768969728).

By making the default (a) something it's now more likely they're trying to use and (b) something that gets to a "would work if you had that board" quickly, it's more likely they'll either get things working, or quickly figure out they're heading in the wrong direction.  (Currently it takes a lot of yak shaving to determine that installing lots of Xilinx things is the wrong direction.)

Capturing a suggestion that @mithro (https://twitter.com/mithro/status/1126988574547558400) and I (https://twitter.com/ewenmcneill/status/1126998432978165760) made independently on Twitter.

mithro:
SGTM!

ewenmcneill:
Closing issue, because the underlying surprise (that prompted this on IRC) got fixed by having *no* default platform -- see https://github.com/timvideos/litex-buildenv/pull/162.   More detail in https://github.com/timvideos/litex-buildenv/pull/138#issuecomment-514872522, https://github.com/timvideos/litex-buildenv/issues/99#issuecomment-514862433, https://github.com/timvideos/litex-buildenv/issues/99#issuecomment-514865663,  https://github.com/timvideos/litex-buildenv/issues/99#issuecomment-514882340, and https://github.com/timvideos/litex-buildenv/issues/99#issuecomment-514884125.

Ewen

Issue #125, opened 2019-04-02
make gateware fails on base target for platform ice40_up5k_b_evn


kamejoko80:
Hello,

I got FPGA placement error when building gateware for platform ice40_up5k_b_evn,  here are my steps below:

export CPU=vexriscv
export TARGET=base
export PLATFORM=ice40_up5k_b_evn
export FIRMWARE=micropython
source scripts/enter-env.sh
make gateware

Error log:

`Info: Device utilisation:
Info:            ICESTORM_LC:  8179/ 5280   154%
Info:           ICESTORM_RAM:    28/   30    93%
Info:                  SB_IO:    16/   96    16%
Info:                  SB_GB:     8/    8   100%
Info:           ICESTORM_PLL:     0/    1     0%
Info:            SB_WARMBOOT:     0/    1     0%
Info:           ICESTORM_DSP:     0/    8     0%
Info:         ICESTORM_HFOSC:     0/    1     0%
Info:         ICESTORM_LFOSC:     0/    1     0%
Info:                 SB_I2C:     0/    2     0%
Info:                 SB_SPI:     0/    2     0%
Info:                 IO_I3C:     0/    2     0%
Info:            SB_LEDDA_IP:     0/    1     0%
Info:            SB_RGBA_DRV:     0/    1     0%
Info:         ICESTORM_SPRAM:     4/    4   100%

Info: Placed 16 cells based on constraints.
Info: Creating initial placement for remaining 8219 cells.
Info:   initial placement placed 500/8219 cells
Info:   initial placement placed 1000/8219 cells
Info:   initial placement placed 1500/8219 cells
Info:   initial placement placed 2000/8219 cells
Info:   initial placement placed 2500/8219 cells
Info:   initial placement placed 3000/8219 cells
Info:   initial placement placed 3500/8219 cells
Info:   initial placement placed 4000/8219 cells
Info:   initial placement placed 4500/8219 cells
Info:   initial placement placed 5000/8219 cells
ERROR: failed to place cell '$abc$60802$auto$blifparse.cc:492:parse_blif$63173_LC' of type 'ICESTORM_LC'
ERROR: Placing design failed.
1 warning, 2 errors
Traceback (most recent call last):
  File "./make.py", line 164, in <module>
    main()
  File "./make.py", line 148, in main
    vns = builder.build(**dict(args.build_option))
[...]

Issue #99, opened 2018-11-13
iceprog isn't provided in conda so icebreaker fails


mithro:
icebreaker programming fails because of missing iceprog.

It should be part of conda.

cr1901:
https://github.com/timvideos/conda-hdmi2usb-packages/blob/master/icestorm/noiceprog.patch Getting rid of this patch should fix the issue. Do you recall why you had it in the first place?

As for why it "worked on my machine":

```
(LX P=tinyfpga_bx.minimal F=micropython) william@xubuntu-dtrain:~/Projects/litex-buildenv$ iceprog
iceprog: missing argument
Try `iceprog --help' for more information.
(LX P=tinyfpga_bx.minimal F=micropython) william@xubuntu-dtrain:~/Projects/litex-buildenv$ which iceprog
/home/william/.local/bin/iceprog
(LX P=tinyfpga_bx.minimal F=micropython) william@xubuntu-dtrain:~/Projects/litex-buildenv$ which icebram
/home/william/Projects/litex-buildenv/build/conda/bin/icebram
(LX P=tinyfpga_bx.minimal F=micropython) william@xubuntu-dtrain:~/Projects/litex-buildenv$ 
```

Looks like the build environment is falling back to my locally-installed version.

cr1901:
@mithro I'm pretty sure we just assume the host provides `libftdi`/`libusb`<sup>1</sup>? In any case both `libftdi`/`libusb` are certainly already required to program arty boards. So I don't see the harm in bringing in `iceprog`.

1.  For now anyway. In my dormant conda-on-windows branch, I ended up building/packaging `libftdi`/`libusb` for Windows users.

mithro:
Yes - `libftdi` / `libusb` shouldn't be required for icestorm. They can be required for iceprog.

ewenmcneill:
After doing `conda update --all` (which downgraded some things) to get a consistent environment after `download-env.sh` and `conda install binutils-riscv32-elf` had done their thing, it's now failing with a different dependency.

My guess is that the *newer* uploads of `iceprog` aren't even being considered due to other dependency problems, and the older uploads of `iceprog` will no longer install because they were built for Python 3.7 (.0?) which is no longer available in Conda.  And/or because it's picking at random -- I can't pin a specific version to install, AFAICT, because they're all the same (upstream) version, just built at different times.

Ewen

```
[...]

PR #138, opened 2019-05-16
Make default platform an open source one (tinyfpga_bx)

ewenmcneill:
Historically the default platform litex-buildenv was the "opsis" video board, because of the origins of litex-buildenv in the HDMI2USB project.  However the opsis requires Xilinx tools installed, which require non-trivial manual steps.  Ease the path for naive users by defaulting to a target that has all open source tooling, so that we can install all tools easily, and thus the user will either quickly reach a useful point, or quickly and painlessly reach a dead end and realise they should set some environment variables.

```
. scripts/enter-env.sh
```
gives `LX P=tinyfpga_bx.minimal F=micropython`

```
export PLATFORM=opsis
. scripts/enter-env.sh
```
gives `LX P=opsis`, as before.

(fixes timvideos/litex-buildenv#134)

Examples:

```
ewen@parthenon:~$ cd /src/fpga/litex-buildenv/
ewen@parthenon:/src/fpga/litex-buildenv$ . scripts/enter-env.sh 
[...]
(LX P=tinyfpga_bx.minimal F=micropython R=default-target-tinyfpga-bx) ewen@parthenon:/src/fpga/litex-buildenv$ make info
              Platform: tinyfpga_bx
                Target: base (default: base)
                   CPU: lm32.minimal (default: lm32)
              Firmware: micropython (default: micropython)
(LX P=tinyfpga_bx.minimal F=micropython R=default-target-tinyfpga-bx) ewen@parthenon:/src/fpga/litex-buildenv$ 
```

```
ewen@parthenon:~$ cd /src/fpga/litex-buildenv/
ewen@parthenon:/src/fpga/litex-buildenv$ export PLATFORM=opsis
ewen@parthenon:/src/fpga/litex-buildenv$ . scripts/enter-env.sh 
[...]
(LX P=opsis R=default-target-tinyfpga-bx) ewen@parthenon:/src/fpga/litex-buildenv$ make info
              Platform: opsis
                Target: video (default: video)
                   CPU: lm32 (default: lm32)
(LX P=opsis R=default-target-tinyfpga-bx) ewen@parthenon:/src/fpga/litex-buildenv$ 
```
```

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -15,7 +15,7 @@ export PYTHON
 SPLIT_REGEX := ^\([^.]*\)\.\?\(.*\)$$
 
 # The platform to run on. It is made up of FPGA_MAIN_BOARD.EXPANSION_BOARD
-DEFAULT_PLATFORM = opsis
+DEFAULT_PLATFORM = tinyfpga_bx
 DEFAULT_PLATFORM_EXPANSION =
 
 ifneq ($(FULL_PLATFORM),)
@@ -80,7 +80,18 @@ ifneq ($(FULL_CPU),)
     endif
 endif
 
+# On our default platform, and CPU, we need a minimal CPU variation to fit,
+# so we default the variant to minimal on the default platform; on any other
+# platform we default to the standard CPU.  We want this to end up in
+# the prompt, so we do not set it as the DEFAULT_CPU_VARIANT, but apply
+# it later.
+#
 CPU ?= $(DEFAULT_CPU)
+ifeq ($(PLATFORM),$(DEFAULT_PLATFORM))
+  ifeq ($(CPU),$(DEFAULT_CPU))
+CPU_VARIANT ?= minimal
+  endif
+endif
 CPU_VARIANT ?= $(DEFAULT_CPU_VARIANT)
 
 ifeq ($(CPU),)
@@ -119,9 +130,10 @@ ifeq ($(TARGET),)
 endif
 export TARGET
 
-FIRMWARE ?= firmware
+DEFAULT_FIRMWARE ?= firmware
+FIRMWARE ?= $(DEFAULT_FIRMWARE)
 ifeq ($(FIRMWARE),)
-    FIRMWARE = firmware
+    FIRMWARE = $(DEFAULT_FIRMWARE)
 endif
 export FIRMWARE
 
@@ -437,6 +449,7 @@ env:
 	@echo "export CPU='$(CPU)'"
 	@echo "export CPU_VARIANT='$(CPU_VARIANT)'"
 	@echo "export FIRMWARE='$(FIRMWARE)'"
+	@echo "export DEFAULT_FIRMWARE='$(DEFAULT_FIRMWARE)'"
 	@echo "export OVERRIDE_FIRMWARE='$(OVERRIDE_FIRMWARE)'"
 	@echo "export PROG='$(PROG)'"
 	@echo "export TARGET_BUILD_DIR='$(TARGET_BUILD_DIR)'"
@@ -463,7 +476,7 @@ info:
 	@echo "                Target: $(TARGET) (default: $(DEFAULT_TARGET))"
 	@echo "                   CPU: $(FULL_CPU) (default: $(DEFAULT_CPU))"
 	@if [ x"$(FIRMWARE)" != x"firmware" ]; then \
-		echo "               Firmare: $(FIRMWARE) (default: firmware)"; \
+		echo "              Firmware: $(FIRMWARE) (default: $(DEFAULT_FIRMWARE))"; \
 	fi
 
 prompt:
--- a/Makefile
+++ b/Makefile
@@ -85,18 +85,7 @@ ifneq ($(FULL_CPU),)
     endif
 endif
 
-# On our default platform, and CPU, we need a minimal CPU variation to fit,
-# so we default the variant to minimal on the default platform; on any other
-# platform we default to the standard CPU.  We want this to end up in
-# the prompt, so we do not set it as the DEFAULT_CPU_VARIANT, but apply
-# it later.
-#
 CPU ?= $(DEFAULT_CPU)
-ifeq ($(PLATFORM),$(DEFAULT_PLATFORM))
-  ifeq ($(CPU),$(DEFAULT_CPU))
-CPU_VARIANT ?= minimal
-  endif
-endif
 CPU_VARIANT ?= $(DEFAULT_CPU_VARIANT)
 
 ifeq ($(CPU),)
--- a/Makefile
+++ b/Makefile
@@ -15,6 +15,11 @@ export PYTHON
 SPLIT_REGEX := ^\([^.]*\)\.\?\(.*\)$$
 
 # The platform to run on. It is made up of FPGA_MAIN_BOARD.EXPANSION_BOARD
+#
+# 2019-05-15 - Default platform is tinyfpga_bx as it is widely available
+#              and has open source tooling; change to icebreaker later in 2019
+# FIXME:       (See https://github.com/timvideos/litex-buildenv/issues/134)
+#
 DEFAULT_PLATFORM = tinyfpga_bx
 DEFAULT_PLATFORM_EXPANSION =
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by ewenmcneill** [comment]
@mithro I *think* this handles both the default and non-default `tinyfpga_bx` platform cases, and the `export PLATFORM=opsis; . scripts/enter-env.sh` case properly, for both `scripts/download-env.sh` and `scripts/enter-env.sh`.   But the `Makefile`s and `scripts/enter-env.sh` / `scripts/download-env.sh` are a rather twisty maze, and we have to do some special things for `tinyfpga_bx` (default `FIRWMARE=micropython`, default `CPU_VARIANT=minimal`) so it's extra twisty.

I'd appreciate it if you could take a look and see if it seems sane to you.

Ewen

PS: Also, we should try to remember to change the default platform to the Icebreaker in about 3 months when those are more generally released (IIRC the *first* early bird orders of Icebreakers shipped earlier this month; but not even all the early bird orders have been assembled yet).

File: Makefile [hunk 1]
```diff
@@ -15,7 +15,7 @@ export PYTHON
 SPLIT_REGEX := ^\([^.]*\)\.\?\(.*\)$$
 
 # The platform to run on. It is made up of FPGA_MAIN_BOARD.EXPANSION_BOARD
-DEFAULT_PLATFORM = opsis
+DEFAULT_PLATFORM = tinyfpga_bx
```
mithro: I think the icebreaker is a better default platform given that we don't have a usb serial UART yet.

Makefile
```diff
@@ -80,7 +80,18 @@ ifneq ($(FULL_CPU),)
...
 endif
 
+# On our default platform, and CPU, we need a minimal CPU variation to fit,
+# so we default the variant to minimal on the default platform; on any other
+# platform we default to the standard CPU.  We want this to end up in
+# the prompt, so we do not set it as the DEFAULT_CPU_VARIANT, but apply
+# it later.
+#
 CPU ?= $(DEFAULT_CPU)
+ifeq ($(PLATFORM),$(DEFAULT_PLATFORM))
```
mithro: I don't think we need to do this... Let me look closer...

targets/tinyfpga_bx/Makefile.mk
```diff
@@ -7,6 +7,8 @@ endif
 # Settings
 DEFAULT_TARGET = base
 TARGET ?= $(DEFAULT_TARGET)
+DEFAULT_FIRMWARE = micropython
```
mithro: I think we want a `DEFAULT_CPU_VARIANT`?

Makefile
ewenmcneill: There are a couple of related problems:

*   The `CPU_VARIANT`s are not very standardised, so the variant that we want depends on the `CPU` as well; and

*   The `CPU` / `CPU_VARIANT` are both set *before* we load anything from the `TARGET` Makefile, which means it's tool late to set a *default* `CPU_VARIANT` in the `TARGET` Makefile

Hence this work around, of setting it in the top level Makefile.  I think I've figured out a (slightly ugly, but hidden away in the `TARGET` Makefile) work around that is sufficient for the TinyFPGA BX / lm32 / minimal case.  I'm testing that now, and will push it if it builds okay.

Ewen

Makefile
ewenmcneill: See rationale in https://github.com/timvideos/litex-buildenv/issues/134 :-)  Ie, almost no one has an Icebreaker yet, including me (I have two on order, including an early bird one; neither has shipped yet), but lots of people, including the person whose problem prompted deciding to change this now, have a TinyFPGA BX.  (And I have a TinyFPGA BX, so I can test it :-) )

I agree the `icebreaker` is a better final target.  I'll add a `FIXME` comment to that effect so we don't forget.

Ewen

targets/tinyfpga_bx/Makefile.mk
ewenmcneill: Aside from the problem that it's too late to set `CPU` related defaults in the `TARGET` Makefile (the [include of the `TARGET` Makefile](https://github.com/timvideos/litex-buildenv/blob/master/Makefile#L115) happens *after* [the `CPU` / `CPU_VARIANT` settings](https://github.com/timvideos/litex-buildenv/blob/7fc049aad997d8deefb5980fc559fce5f2170675/Makefile#L57-L84)), there's also the problem that if we could set `DEFAULT_CPU_VARIANT` then `minimal` no longer appears in the prompt (since the `CPU_VARIANT` is suppressed from the prompt if `CPU_VARIANT = DEFAULT_CPU_VARIANT`), and I think it's useful that it appears in the prompt at least in this case (as it's a really cut down CPU which is *slow*).

Besides, *not* setting `DEFAULT_CPU_VARIANT` is bug-for-bug compatible with what happens before.

Ewen

Makefile
mithro: I ran into this problem (via https://github.com/timvideos/litex-buildenv/issues/125) and fixed it see https://github.com/enjoy-digital/litex/pull/175 and https://github.com/enjoy-digital/litex/issues/159 . 

We just need to pull in the new litex.

targets/tinyfpga_bx/Makefile.mk
mithro: Can we just move the CPU stuff to after the `TARGET` Makefile?

Makefile
ewenmcneill: I *think* you're addressing the second test (on the `CPU` being the `DEFAULT_CPU`), rather than the `PLATFORM` test (since the `PLATFORM` test is gone now it's in the target Makefile.  But even with the names standardised (which is a great step forward!), it's possible that, eg, only lm32.minimal fits, but vexrisc.SOMETHING (else larger) will fit.  So I think the default variants required are still somewhat CPU specific.  (Ideally we'd probably set `DEFAULT_FULL_CPU`, and then split that apart.)

Ewen

targets/tinyfpga_bx/Makefile.mk
ewenmcneill: *Maybe* :-)

I did think about that, and in general I think it's a good idea.  But I certainly don't have time right now to audit whether there's anything that currently relies on CPU being set before the TARGET Makefile.  (And the `target/tinyfpga_bx/Makefile.mk` now relies on that... :-)  But that one is at least easily fixable by setting a default CPU there too.)

So this work around seemed like a less invasive change.   Feel free to look into the larger refactoring.... :-)

Ewen

Makefile
ewenmcneill: If we wait long enough, my first icebreaker should arrive (in theory the last of the earlybird orders ship this month, so I guess mine should arrive in the next 2-3 weeks... :-) ).  At which point I could test defaulting to the icebreaker and do that instead.  (I think we possibly still want some of the other changes in this pull request, to have useful TinyFPGA BX defaults.)

Ewen

ewenmcneill: FTR, I've just learned that my iCEBreaker early bird order has been held back "to ship with the main production batch" (apparently it was a mistake to order a second one from the main batch, even though I paid separate shipping for each one :-( ).  

So "waiting long enough" probably means another couple of months.

Ewen

ewenmcneill:
Closing pull request because based on discussion (https://github.com/timvideos/litex-buildenv/issues/99#issuecomment-514862433 and https://github.com/timvideos/litex-buildenv/issues/99#issuecomment-514865663) the entire issue seems to have been rendered moot by having *no* default platform, which seems to have been merged a few weeks ago in https://github.com/timvideos/litex-buildenv/pull/162.  So there doesn't seem to be any point keeping this pull request open.

Ewen

PS: FTR, my iCEBreaker boards finally arrived (yesterday).

```
ewen@parthenon:/src/fpga/litex-buildenv$ . scripts/enter-env.sh 
             This script is: /src/fpga/litex-buildenv/scripts/enter-env.sh
         Firmware directory: /src/fpga/litex-buildenv
         Build directory is: /src/fpga/litex-buildenv/build
     3rd party directory is: /src/fpga/litex-buildenv/third_party

Unknown platform ''

Valid platforms are:
 * arty
 * atlys
 * basys3
 * cmod_a7
 * galatea
 * ice40_hx8k_b_evn
 * ice40_up5k_b_evn
 * icebreaker
 * mimas_a7
 * mimasv2
 * minispartan6
 * neso
 * nexys_video
 * opsis
 * pipistrello
 * saturn
 * sim
 * tinyfpga_bx
 * upduino_v1
 * waxwing
ewen@parthenon:/src/fpga/litex-buildenv$ 
```

mithro:
@ewenmcneill Felt like not having a default platform makes it easier to not accidentally get the wrong set of tools.

Issue #131, opened 2019-05-03
Bootstrap instructions are debian specific


mballance:
@mithro says "Everything else should work but the bootstrap is debian specific."

mithro:
Work in progress at https://github.com/timvideos/litex-buildenv/pull/132

mithro:
This has been fixed in #132. 

 * The `download-env-root.sh` script has been renamed to `debian-setup.sh`.
 * `bootstrap.sh` script only calls `debian-setup.sh` on Debian / Ubuntu systems. Otherwise it checks for the few things needed and tells you to install them.

TomKeddie:
@mithro  I think the grep -v is wrong, the -v at https://github.com/timvideos/litex-buildenv/blob/master/scripts/bootstrap.sh#L42 the logic I don't think we want this here.  On archlinux I see it try to run the debian script.
```
bash: line 43: ./scripts/debian-setup.sh: No such file or directory
```

Also when running bootstrap via curl as per [wiki](https://github.com/timvideos/litex-buildenv/wiki/Bootstrap) the other scripts won't be present.

mithro:
@TomKeddie You seem to be correct, it should be `grep -q` -- Do you want to submit your first pull request for fixing that?

mithro:
Regarding;
> Also when running bootstrap via curl as per [wiki](https://github.com/timvideos/litex-buildenv/wiki/Bootstrap) the other scripts won't be present.

Notice the script clones the github repository and then runs the scripts? See the following lines;

https://github.com/timvideos/litex-buildenv/blob/7fc049aad997d8deefb5980fc559fce5f2170675/scripts/bootstrap.sh#L33-L44

TomKeddie:
> Notice the script clones the github repository and then runs the scripts? See the following lines;

I see the problem now, the bootstrap instructions pull bootstrap.sh from litex-buildenv but it clones HDMI2USB-litex-firmware which doesn't have the changes, would you prefer the wiki point to HDMI2USB-litex-firmware for bootstrap?

mithro:
@TomKeddie I think LiteX-BuildEnv is probably the right default?

TomKeddie:
@mithro I'm not sure what the grand plan is but I think should should be consistent, should I do a PR for bootstrap.sh then?

mithro:
@TomKeddie We are making it up as we go along :-)

[...]

PR #142, opened 2019-05-21
bootstrap: fix logic inversion in platform detection

TomKeddie:
Fix issues with debian detection mentioned in https://github.com/timvideos/litex-buildenv/issues/131


mithro:
Thanks for the pull request!

PR #143, opened 2019-05-22
scripts : update bootstrap.sh to use new repo name

TomKeddie:
As noted in https://github.com/timvideos/litex-buildenv/issues/131

### DIFF
--- a/scripts/bootstrap.sh
+++ b/scripts/bootstrap.sh
@@ -10,7 +10,7 @@ if [ -z "$GITHUB_USER" ]; then
   GITHUB_USER=timvideos
 fi
 if [ -z "$GITHUB_REPO" ]; then
-  GITHUB_REPO=HDMI2USB-litex-firmware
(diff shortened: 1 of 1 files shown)


scripts/bootstrap.sh
```diff
@@ -10,7 +10,7 @@ if [ -z "$GITHUB_USER" ]; then
   GITHUB_USER=timvideos
 fi
 if [ -z "$GITHUB_REPO" ]; then
-  GITHUB_REPO=HDMI2USB-litex-firmware
+  GITHUB_REPO=LiteX-BuildEnv
```
mithro: I think this should be lowercase?

TomKeddie: It works fine either way, I used the case you used in the bug, but am indifferent, changed to lower case.

mithro:
@TomKeddie Thanks for the pull request!

PR #147, opened 2019-06-12
travis: test zephyr script and save built binary

mateusz-holenko:
This adds testing of `build-zephyr.sh` script on travis CI and saving the generated binary to prebuilt repo.


mithro:
Can you add the Travis config to run the Zephyr build?

mateusz-holenko:
@mithro I noticed that CI has failed, but is it related to my changes? Is it possible to start it again?

mithro:
@mateusz-holenko Just push an update

mateusz-holenko:
There is a problem with getting `zephyr_sdk` (~ 770MB). The connection speed is low, so it takes too long to download it and the connection is dropped in the meantime.

Is there a simple way to embedd this into the image so it does not have to be downloaded every time?

mithro:
@mateusz-holenko I think we will have to mirror it somewhere that is faster to download from. 770mb should take a couple of seconds...

scripts/download-env.sh
```diff
@@ -486,6 +486,14 @@ if [ "$FIRMWARE" = "zephyr" ]; then
 	echo "Installing pykwalify (pyhton module)"
 	pip install pykwalify
 	check_import pykwalify.core
+
+	# cmake for building Zephyr
+	# Zephyr requires cmake 3.13.1 or higher
+	# installing it via pip provides 3.14.4
+	# at the time of writing this comment
+	echo
+	echo "Installing cmake (pyhton module)"
```
mithro:
```suggestion
	echo "Installing cmake (python module)"
```

scripts/enter-env.sh
```diff
@@ -344,6 +344,34 @@ check_version ${CPU_ARCH}-elf-gcc $GCC_VERSION || return 1
 #
 #check_version ${CPU_ARCH}-elf-gdb $GDB_VERSION
 
+# Cmake Version
+###############
+if [ "$FIRMWARE" = "zephyr" ]; then
+	set -x
+	# Zephyr requires cmake 3.13.1 or higher
+	CMAKE_VERSION=$(cmake --version | head -n1 | cut -d' ' -f3)
```
mithro: There is a check version function?

mateusz-holenko:
The proper place for hosting Zephyr SDK needs to be decided before merging this one.

The current one is for testing purposes.

scripts/enter-env.sh
mateusz-holenko: `check_version` function verifies an exact version - here I check for version 3.13.1 **or higher**

I agree that it adds complexity, so the other solution I see is to force a 3.14.4 version in pip and simply check here for this version using `check_version`. What do you think @mithro?

mateusz-holenko: I fixed the cmake version to 3.14.4

mateusz-holenko:
@mithro Travis CI is constantly failing on installing python using conda. I had the same problems yesterday on my local machine, but today it works fine.

I push-forced the branch today few times, but with no success. Do you have an idea how to solve this issue?

mithro:
Looks like something is screwed up at conda's end?

mithro:
Could also have something to do with our package pins?

mithro:
Blocked on https://github.com/conda/conda/issues/8843

scripts/download-env.sh
```diff
@@ -482,6 +482,16 @@ if [ "$FIRMWARE" = "zephyr" ]; then
 	conda install -y $CONDA_FLAGS pyyaml
 	check_import yaml
 
+	# gperf for Zephyr SDK
```
mithro: These should be check in `enter-env.sh`...

PR #154, opened 2019-06-18
Renode boot from flash

mateusz-holenko:
This PR adds two features to Renode emulation:

* support for shadow addresses,
* ability to boot from flash (no netboot needed).


mithro:
Does renode have proper models of the nor spiflash parts?

mateusz-holenko:
We have a model of Micron [MT25Q](https://www.micron.com/products/nor-flash/serial-nor-flash/mt25q).

In order to make things simple we usually model read-only XIP flashes as a memory that is loaded directly from Renode's monitor or script.

mithro:
@mateusz-holenko - I think we want to support both modes? IE Writing to the SPI flash is a good thing you want to test.

mateusz-holenko:
@mithro Sure.
I checked and this would require to add an spi flash controller (the one that does bitbanging) model to Renode, but it's not a problem.

Is there any firmware I could run to test if everything works as expected?
I found the ["driver" code](https://github.com/enjoy-digital/litex/blob/master/litex/soc/software/libbase/spiflash.c), but I see no place where any of `write_to_flash`, `write_to_flash_page` or `erase_flash_sector` is used.

mateusz-holenko:
Since I couldn't find any firmware that writes to flash I decided to add a new command to LiteX BIOS ([PR](https://github.com/enjoy-digital/litex/pull/204) waiting).

I created a LiteX SPI flash controller model in Renode (currently under review) and once it's merged to master I will update this PR to use it in dynamically generated platform.

mithro:
Blocked on https://github.com/conda/conda/issues/8843

mateusz-holenko:
@mithro Travis is finally green. Is the current version fine for you?

PR #164, opened 2019-07-08
platform/icebreaker: port rgb led driver from fomu

TomKeddie:


targets/icebreaker/base.py
```diff
@@ -57,6 +58,55 @@ def __init__(self, platform):
         self.specials += AsyncResetSynchronizer(self.cd_por, reset)
 
 
+class _SBLED(Module, AutoCSR):
```
mithro: Let's move this module under the gateware directory so I can be shared between all iCE40 platforms?

mithro: FYI - I did this and pushed to your branch, will merge if things go green.

mithro:
I don't know why Travis isn't running here....

PR #168, opened 2019-07-10
scripts : fix download-env/enter-env platform detection

TomKeddie:
to work outside litex-buildenv dir


mithro:
Will merge if goes green.

PR #176, opened 2019-07-24
Fix endianness

mateusz-holenko:
This PR fixes two problems with endianness for little endian CPUs (like VexRiscv):

* liteeth was configured to big endian and it was not possible to netboot in LiteX bios on VexRiscv,
* uip was configured to big endian and it was not possible to telnet into the firmware on VexRiscv.

### DIFF
--- a/targets/arty/net.py
+++ b/targets/arty/net.py
@@ -31,7 +31,7 @@ class NetSoC(BaseSoC):
             platform.request("eth_clocks"),
             platform.request("eth"))
         self.submodules.ethmac = LiteEthMAC(
-            phy=self.ethphy, dw=32, interface="wishbone")
+            phy=self.ethphy, dw=32, interface="wishbone", endianness=self.cpu.endianness)
         self.add_wb_slave(mem_decoder(self.mem_map["ethmac"]), self.ethmac.bus)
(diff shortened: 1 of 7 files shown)


targets/arty/net.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, platform, *args, **kwargs):
             platform.request("eth_clocks"),
             platform.request("eth"))
         self.submodules.ethmac = LiteEthMAC(
```
mithro: I don't quite understand why LiteEthMAC needs to understand the cpu endianness? However it looks like that is something from upstream, not something you invented?

enjoy-digital: The MAC converts 32-bit writes/reads to/from 8-bit streams so needs to know the endiannness to know in which order reading/storing the streams.

mateusz-holenko: I have detected the problems when testing the firmware on HW. Without setting the proper endianess, data read from `liteeth` buffers had each 4-byte chunk reverted, causing the software to fail to detect the proper type of packet and drop it in a process.

After realising what was going on, I tested the pure LiteX platform and it worked fine. Diffing them pointed to the `endianness` property (not set in LiteX-BuildEnv).

PR #177, opened 2019-07-25
scripts: Always update conda to pinned version.

mithro:
Potentially fixes @ewenmcneill ran into in #99.


ewenmcneill:
@mithro That seems like a reasonable approach to try to "catch up" existing installs to the new approach of also pinning on the conda version (to avoid problems with newer conda having dependency conflicts with packages we build ourselves).

Ewen

PR #167, opened 2019-07-10
gateware/ice40: add mux stage and direct register drive to LED block

TomKeddie:
As requested in https://github.com/im-tomu/foboot/issues/20

PR #180, opened 2019-07-26
ice40/led : add mux stage and direct register drive to LED block 

TomKeddie:
replaces https://github.com/timvideos/litex-buildenv/pull/167


TomKeddie:
@mithro do we still want this?

mithro:
I think so!

It is unclear to me why I didn't merge it?

TomKeddie:
Cool, I'll t at it some more and then rebase.

PR #190, opened 2019-08-05
Add Azure Pipelines config

auscompgeek:
Still need to:

- [ ] Update the CI scripts to use the Azure Pipelines environment variables
- [ ] Investigate the equivalent of `allow_failures`


auscompgeek:
Here is the pipeline running on my account:

https://auscompgeek.visualstudio.com/litex-buildenv/_build?definitionId=1

auscompgeek:
Note to self:

https://docs.microsoft.com/en-us/azure/devops/pipelines/migrate/from-travis?view=azure-devops#predefined-variables

auscompgeek:
Heh, a couple of the builds need bc or flex, which aren't installed in the image by default.

auscompgeek:
It looks like we might also be able to talk to Microsoft to bump the free tier limits: https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/hosted?view=azure-devops#capabilities-and-limitations

auscompgeek:
Anyone have any idea why the lm32 opsis hdmi2usb build would be flaky? https://dev.azure.com/auscompgeek/litex-buildenv/_build/results?buildId=8

azure-pipelines.yml
```diff
@@ -0,0 +1,422 @@
+pool:
```
mithro: Is it possible for this to be `.github/azure-pipelines.yml`?

auscompgeek: Apparently it's _possible_, but very non-standard, and apparently requires a counterintuitive process to set up. https://stackoverflow.com/questions/55614307/can-you-change-the-location-of-azure-pipelines-yaml-in-azure-devops

It looks like CPython have their config in `.azure-pipelines/ci.yml`, so may be worth looking into how exactly that's set up there.

PR #204, opened 2019-10-30
Allow to pass additional arguments to Renode

mateusz-holenko:
This PR changes the way Renode is started by the `./scripts/build-renode.sh` script.

Instead of just passing a generated `resc` script to Renode, a series of `-e` switches is used. It allows user to pass own additional arguments before starting the simulation.

*Note:* currently the generated script contains a `start` command at the end so the last `-e s` is not necessary, but once this PR is merged I will remove the final `start` from the generator. Proceeding in this order is to ensure that this script works during the transition period.


mithro:
Please rebase on top of master (should fix travis issue).

PR #223, opened 2019-11-08
platforms: add pano logic g2 platform

piotr-binkowski:
Aim of this PR is to add initial support for [Pano Logic Zero Client G2](https://github.com/tomverbeure/panologic-g2).
Initial version supports only a basic set of features:
* UART exposed on DVI DDC pins
* Single DDR2 chip

With this configuration it is possible to run Linux built using `./scripts/build-linux.sh`, only catch is that it must be transferred over a serial connection.

Additionally I'm not sure how should I implement the `gateware-load` target and similar because the device has only a JTAG header and no built-in programmer

### DIFF
--- a/platforms/pano_logic_g2.py
+++ b/platforms/pano_logic_g2.py
@@ -1,0 +1,185 @@
+# Support for the Pano Logic Zero Client G2
+# More information about the board and the reverse engineering results
+# can be found here https://github.com/tomverbeure/panologic-g2
+
+from litex.build.generic_platform import *
+from litex.build.xilinx import XilinxPlatform, iMPACT
+
+# IOs ----------------------------------------------------------------------------------------------
+
+_io = [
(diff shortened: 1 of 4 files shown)


mithro:
Can you add the platform to the Travis target? Will probably need to be `allow_failure` for now as the Xilinix toolchain probably doesn't support the part used so far.

platforms/pano_logic_g2.py
```diff
@@ -0,0 +1,183 @@
+# Support for the Pano Logic Zero Client G2
```
mithro: Please add a bunch of comments to the top of this file on where to find more information.

mithro:
FYI - https://github.com/tomverbeure/panologic-g2

piotr-binkowski:
I've added information on where to find more about the Pano Logic G2 and added a Travis target for this board

mithro:
@john-k

PR #230, opened 2019-11-12
scripts/build-linux.sh: add option to build rootfs and devicetree from scratch

piotr-binkowski:
Currently the script uses prebuilt rootfs and devicetree for VexRiscv configuration, this PR aims to add option to build Linux using Buildroot from [linux-on-litex-vexriscv](https://github.com/litex-hub/linux-on-litex-vexriscv).
This also fixes the issue with LiteX configuration not matching the prebuilt devicetree which for example surfaced in #223 where board has 64MiB of DRAM but devicetree specifies only 32MiB

By default the old method of building Linux is used, to use Buildroot first export `BUILD_BUILDROOT=1`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,5 @@ third_party/qemu-litex
 third_party/micropython
 third_party/linux
 third_party/litex-devicetree
+third_party/buildroot
+third_party/linux-on-litex-vexriscv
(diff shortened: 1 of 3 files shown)


mithro:
We shouldn't need the linux-on-litex repository. Instead push the patches into litex-hub/linux and just clone that.

We should also always be generating the device tree file.

The buildroot change looks good!

PR #238, opened 2019-11-21
Add NeTV2 support and HDMI2PCIe

piotr-binkowski:
This PR aims to add support for NeTV2 and a design that allows to stream in and out video data using PCIe interface. At this point the initial LiteX design and firmware are part of this PR but I will soon add a Linux driver for use with the design


piotr-binkowski:
Right, there is a fair bit of code duplication with the current solution.
I will look into reusing the existing firmware.

piotr-binkowski:
I've added initial support for transferring video data over PCIe to the existing firmware. I had to adjust framebuffer offsets so that driver can access the HDMI buffers using 32MB BAR0 as this initial design still uses BARs to access the card memory. Additonally I added a simple python app that uses the Linux driver

piotr-binkowski:
I've updated the sample app so it can handle UYVY output from Imagemagick without any need for byte reversing and similarly its output can now be used without reversing byte order

firmware/framebuffer.h
```diff
@@ -22,16 +22,17 @@
...
  *
  *  0x01000000 - Pattern Buffer - Frame Buffer n
+ *  0x04000000 - PCIe Buffer - Frame Buffer n
  *
  * Each input then has 3 frame buffers spaced like this;
  *  // HDMI Input 0
- *  0x02000000 - HDMI Input 0 - Frame Buffer n
- *  0x02040000 - HDMI Input 0 - Frame Buffer n+1
- *  0x02080000 - HDMI Input 0 - Frame Buffer n+2
+ *  0x05000000 - HDMI Input 0 - Frame Buffer n
```
mithro: Why have these moved?

.travis.yml
```diff
@@ -82,6 +82,7 @@ env:
    - C=mor1kx           TC="ise"       P=opsis             T="base net"
    - C=mor1kx           TC="ise"       P=pipistrello       T="base"
    # VexRISCV
+   - C=vexriscv         TC="vivado"    P=netv2             T="base"
```
mithro: NeTV2 should probably have a net target as well?

firmware/ci.c
```diff
@@ -647,6 +647,10 @@ static void video_matrix_list(void)
 #ifdef CSR_HDMI_IN1_BASE
 	wprintf("input1 (1): %s\n", HDMI_IN1_MNEMONIC);
 	wputs(HDMI_IN1_DESCRIPTION);
+#endif
+#ifdef CSR_PCIE_PHY_BASE
+	wprintf("pcie:\n");
```
mithro: Should probably add a short code for PCIe?

firmware/ci.c
```diff
@@ -669,7 +673,7 @@ static void video_matrix_list(void)
 
 static void video_matrix_connect(int source, int sink)
 {
-	if(source >= 0 && source <= VIDEO_IN_PATTERN)
+	if(source >= 0 && source <= VIDEO_IN_PCIE)
```
mithro: We should add a `VIDEO_IN_MAX` or similar constant.

firmware/framebuffer.h
```diff
@@ -40,15 +41,19 @@
  *
  */
 #define FRAMEBUFFER_OFFSET		0x01000000
```
mithro: This should probably be based on the value of `MAIN_RAM_OFFSET` or similar?

Could actually use the linker script to place these buffers...

mithro:
My documents ghjMn mjkxnZ

firmware/ci.c
piotr-binkowski: `p` is used by pattern so maybe `e`?

firmware/framebuffer.h
piotr-binkowski: Input buffers and PCIe buffer were moved so that PCIe-Wishbone bridge can access the data inside them.
Bridge can access 32M of memory starting at 0x44000000.

software/hdmi2pcie/kernel/main.c
```diff
@@ -0,0 +1,385 @@
+/*
+ * NeTV2 PCIe driver
```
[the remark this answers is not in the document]
piotr-binkowski: right, good catch

piotr-binkowski:
@mithro I think you closed this PR by accident

mithro:
@piotr-binkowski You are correct -- I think my pocket actually closed it....

firmware/framebuffer.h
piotr-binkowski: I think that those offsets are already internally based on `MAIN_RAM_BASE` position because when CPU accesses those buffers it [adds](https://github.com/timvideos/litex-buildenv/blob/master/firmware/pattern.c#L142) the `MAIN_RAM_BASE` to the offset.
And DMA controllers [take](https://github.com/timvideos/litex-buildenv/blob/master/firmware/processor.c#L726) only the offset value which means that they already use addresses relative to main memory

firmware/processor.h
```diff
@@ -47,13 +47,15 @@ enum {
 enum {
 	VIDEO_IN_HDMI_IN0=0,
 	VIDEO_IN_HDMI_IN1,
-	VIDEO_IN_PATTERN
+	VIDEO_IN_PATTERN,
+	VIDEO_IN_PCIE
```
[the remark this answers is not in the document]
piotr-binkowski: done

firmware/ci.c
piotr-binkowski: done

.travis.yml
piotr-binkowski: In theory yes, but at this point NeTV2 lacks a `net.py` file that instantiates the network interface. I focused more on PCIe side of things in this PR and I didn't test the network interface

firmware/ci.c
piotr-binkowski: I've added `e` as a short code for PCIe input

software/hdmi2pcie/kernel/main.c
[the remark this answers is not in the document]
piotr-binkowski: done

piotr-binkowski:
@mithro please take a look
I think that I addressed all the comments and after CI finishes this can be merged

.travis.yml
piotr-binkowski: I added a network target for NeTV2 and modified the `.travis.yml` to test it.

piotr-binkowski:
I've updated the linux driver to now work using a V4L2 API and added a module that changes the endianess of data coming to and from the PCIe bridge because transferred YUV frames had wrong byte order and previously this conversion was performed in software on a host PC

platforms/netv2.py
```diff
@@ -0,0 +1,222 @@
...
+    # hdmi out
+    ("hdmi_out", 0,
+        Subsignal("clk_p", Pins("W19"), Inverted(), IOStandard("TMDS_33")),
+        Subsignal("clk_n", Pins("W20"), Inverted(), IOStandard("TMDS_33")),
+        Subsignal("data0_p", Pins("W21"), IOStandard("TMDS_33")),
+        Subsignal("data0_n", Pins("W22"), IOStandard("TMDS_33")),
+        Subsignal("data1_p", Pins("U20"), IOStandard("TMDS_33")),
+        Subsignal("data1_n", Pins("V20"), IOStandard("TMDS_33")),
+        Subsignal("data2_p", Pins("T21"), IOStandard("TMDS_33")),
+        Subsignal("data2_n", Pins("U21"), IOStandard("TMDS_33"))
```
mithro: No I2C on `hdmi_out`s?

platforms/netv2.py
```diff
@@ -0,0 +1,222 @@
...
+        Subsignal("data0_p", Pins("E22"), IOStandard("TMDS_33"), Inverted()),
+        Subsignal("data0_n", Pins("D22"), IOStandard("TMDS_33"), Inverted()),
+        Subsignal("data1_p", Pins("C22"), IOStandard("TMDS_33"), Inverted()),
+        Subsignal("data1_n", Pins("B22"), IOStandard("TMDS_33"), Inverted()),
+        Subsignal("data2_p", Pins("B21"), IOStandard("TMDS_33"), Inverted()),
+        Subsignal("data2_n", Pins("A21"), IOStandard("TMDS_33"), Inverted()),
+    ),
+]
+
+_hdmi_infos = {
```
mithro: Can you fix these?

software/hdmi2pcie/kernel/main.c
```diff
@@ -0,0 +1,744 @@
...
+
+	chan->dir = dir;
+
+	ret = v4l2_device_register(&pdev->dev, &chan->v4l2_dev);
+	if (ret) {
+		dev_err(&pdev->dev, "v4l2_device_register failed, ret=%d\n", ret);
+		return ret;
+	}
+
+	// Point to last 4 bytes of a 16MiB area
```
mithro: I think you want to say something like "The last 4 bytes contain the the frame pointer buffer index. This is the index which the FPGA firmware is currently writing to".

software/hdmi2pcie/kernel/main.c
```diff
@@ -0,0 +1,744 @@
...
+		dev_err(dev, "Failed to set DMA mask\n");
+		goto fail4;
+	};
+
+	ret = pci_enable_msi(pdev);
+	if (ret) {
+		dev_err(dev, "Failed to enable MSI\n");
+		goto fail4;
+	}
+
```
mithro: Shouldn't this technically register 2 in and 2 output channels? Can you add a TODO here?

software/hdmi2pcie/kernel/main.c
```diff
@@ -0,0 +1,744 @@
...
+	ret = hdmi2pcie_register_video_dev(pdev, &priv->channels[0], HDMI2PCIE_DIR_IN);
+	ret += hdmi2pcie_register_video_dev(pdev, &priv->channels[1], HDMI2PCIE_DIR_OUT);
+	if (ret) {
+		dev_err(dev, "Failed to register V4L2 device");
+		goto fail5;
+	}
+
+	return 0;
+
+fail5:
```
mithro: Can you please use better names than `fail5` -- something like `fail_before_register_video_dev` might be better...

targets/netv2/hdmi2pcie.py
```diff
@@ -0,0 +1,133 @@
...
+        self.submodules.pcie_phy = S7PCIEPHY(platform, platform.request("pcie_x1"), bar0_size=32*1024*1024)
+        platform.add_false_path_constraints(
+            self.crg.cd_sys.clk,
+            self.pcie_phy.cd_pcie.clk)
+
+        # pcie endpoint
+        self.submodules.pcie_endpoint = LitePCIeEndpoint(self.pcie_phy)
+
+        # pcie wishbone bridge
+        self.submodules.pcie_bridge = LitePCIeWishboneBridge(self.pcie_endpoint, lambda a: 1, shadow_base=0x40000000)
```
mithro: What is the `lambda a: 1` here?

mithro:
Just some minor comments and then good to merge!

software/hdmi2pcie/kernel/main.c
piotr-binkowski: `goto` labels should now use more meaningful names

targets/netv2/hdmi2pcie.py
piotr-binkowski: `lambda a: 1` is used as a minimal address decoder ie. we have one bridge and every access that comes over PCIe is meant for this WB bridge

software/hdmi2pcie/kernel/main.c
piotr-binkowski: TODO added

software/hdmi2pcie/kernel/main.c
piotr-binkowski: I've updated the comment to say more about what this variable is used for

platforms/netv2.py
piotr-binkowski: I've added I2C to HDMI output 1 but from what I see on the schematics HDMI out 0 and in 0 are sharing an I2C bus and pins

platforms/netv2.py
piotr-binkowski: I've added basic information on where each connector is located

piotr-binkowski:
@mithro please take a look, I think that I've fixed all the remaining minor issues

platforms/netv2.py
mithro: @piotr-binkowski IIRC I thought there was a set of jumpers which directly connected HDMI Out 0 to HDMI In 0 or let you control them independently? Let's move to a different PR?

targets/netv2/hdmi2pcie.py
mithro: Please add a comment for this.

PR #244, opened 2019-11-26
Add Fomu support

piotr-binkowski:
The aim of this PR is to add support for Fomu.
Initial design uses the Foboot configuration/firmware which requires special handling because it needs a custom BIOS, due to that at this point it is necessary to first run `make foboot-bios` to build the replacement BIOS before building other targets like `gateware`
It is also required to manually patch foboot with `firmware/foboot/putchar-fix.patch`

### DIFF
--- a/targets/fomu/Makefile.mk
+++ b/targets/fomu/Makefile.mk
@@ -1,0 +1,80 @@
+# fomu targets
+
+ifneq ($(PLATFORM),fomu)
+	$(error "Platform should be ice40-up5k-uwg30 or ice40-up5k-sg48 when using this file!?")
+endif
+
+PPATH := $(PYTHONPATH)
+export PYTHONPATH=$(PPATH):$(TARGET_BUILD_DIR)/../../third_party/valentyusb/
+
+# Settings
+DEFAULT_TARGET = base
+TARGET ?= $(DEFAULT_TARGET)
+BAUD ?= 115200
+
+# Image
+image-flash-$(PLATFORM):
+	cp $(IMAGE_FILE) $(IMAGE_FILE).dfu
+	dfu-suffix --pid 1209 --vid 5bf0 --add $(IMAGE_FILE).dfu
+	dfu-util -D $(IMAGE_FILE).dfu
+
+# Gateware
+gateware-load-$(PLATFORM):
+	@echo "Fomu doesn't support loading, use the flash target instead."
+	@echo "make gateware-flash"
+	@false
+
+# As with Mimasv2, if the user asks to flash the gateware only, the BIOS must
+# be sent as well (because the BIOS is too big to fit into the bitstream).
+#
+# We have to pre-calculate what the image file will end up being, as we are
+# included before it has been defined (to get the default target), so we'll
+# end up comparing with an empty string/older value from the environment.
+#
+GATEWARE_BIOS_FILE = $(TARGET_BUILD_DIR)/image-gateware+bios+none.bin
+IMAGE_FW_FILE      = $(TARGET_BUILD_DIR)/image-gateware+bios+$(FIRMWARE).bin
+
+gateware-flash-$(PLATFORM): $(GATEWARE_BIOS_FILE)
+	cp $(GATEWARE_BIOS_FILE) $(GATEWARE_BIOS_FILE).dfu
+	dfu-suffix --pid 1209 --vid 5bf0 --add $(GATEWARE_BIOS_FILE).dfu
+	dfu-util -D $(GATEWARE_BIOS_FILE).dfu
+
+# To avoid duplicating the mkimage.py call here, if the user has not
+# already built a image-gateware+bios+none.bin, we call make recursively
+# to build one here, with the FIRMWARE=none override.
+#
+ifneq ($(GATEWARE_BIOS_FILE),$(IMAGE_FW_FILE))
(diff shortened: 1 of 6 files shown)


rtl/2-stage-1024-cache.v_toplevel_RegFilePlugin_regFile.bin
```diff
@@ -0,0 +1,32 @@
+00000000000000000000000000000000
```
mithro: This CL shouldn't be adding an `rtl` directory. All `rtl` is under the `gateware` directory.

targets/fomu/Makefile.mk
```diff
@@ -0,0 +1,79 @@
...
+BAUD ?= 115200
+
+MAKE_LITEX_EXTRA_CMDLINE += "--no-compile-software"
+
+FOBOOT_BIN = $(TARGET_BUILD_DIR)/../../rtl/foboot.bin
+FOBOOT_DIR = $(TARGET_BUILD_DIR)/../../third_party/foboot/sw
+
+# Image
+image-flash-$(PLATFORM):
+	iceprog $(IMAGE_FILE)
```
mithro: `iceprog` isn't use with Fomu -- you want `dfu-util`

rtl/2-stage-1024-cache.v_toplevel_RegFilePlugin_regFile.bin
piotr-binkowski: I've moved files from `rtl` directory to `gateware`

piotr-binkowski: I've moved files from rtl directory to gateware

targets/fomu/Makefile.mk
piotr-binkowski: At this point the `base` target for Fomu builds the whole bootloader image (gateware + foboot as bios). I think that loading it must be done using `iceprog` and a direct flash connection. Or is it possible to update the bootloader using `dfu-util`?

piotr-binkowski:
@mithro please take a look
I think that I addressed all of your comments and if everything is all right then this can be merged

targets/fomu/Makefile.mk
mithro: It is possible to update the bootloader via dfu-util

targets/fomu/base.py
```diff
@@ -0,0 +1,451 @@
...
+        self.comb += self.cd_usb_12.clk.eq(clk12)
+
+        self.sync.por += \
+            If(reset_delay != 0,
+                reset_delay.eq(reset_delay - 1)
+            )
+        self.specials += AsyncResetSynchronizer(self.cd_por, self.reset)
+
+
+class SBLED(Module, AutoCSR):
```
mithro: This should be in `gateware` directory. I believe there is a version at https://github.com/timvideos/litex-buildenv/blob/master/gateware/ice40.py

targets/fomu/base.py
```diff
@@ -0,0 +1,451 @@
...
+            # o_LEDDON = led_is_on, # Indicates whether LED is on or not
+            # i_LEDDRST = ResetSignal(), # This port doesn't actually exist
+            o_PWMOUT0 = ledd_value[0],
+            o_PWMOUT1 = ledd_value[1],
+            o_PWMOUT2 = ledd_value[2],
+            o_LEDDON = Signal(),
+        )
+
+
+class SBWarmBoot(Module, AutoCSR):
```
mithro: Should be in https://github.com/timvideos/litex-buildenv/blob/master/gateware/ice40.py

targets/fomu/base.py
```diff
@@ -0,0 +1,451 @@
...
+            touch1.oe.eq(self.oe.storage[0]),
+            touch2.oe.eq(self.oe.storage[1]),
+            touch3.oe.eq(self.oe.storage[2]),
+            touch4.oe.eq(self.oe.storage[3]),
+
+            self.i.status.eq(Cat(touch1.i, touch2.i, touch3.i, touch4.i))
+        ]
+
+
+class PicoRVSpi(Module, AutoCSR):
```
mithro: Should be in `gateware`.

mithro:
The aim here is to have a SoC you can load into the "user" part of the Fomu. This is the image that boots *after* the bootloader. See the "Migen and LiteX" section of https://workshop.fomu.im/#migen-and-litex

mithro:
Could you please rebase this pull request?

mithro:
@mateusz-holenko -- IIRC Could you work with @piotr-binkowski and @ewenmcneill to get this pull request merged?

piotr-binkowski:
I've rebased this and moved module definitions to `gateware` but I still need to update this so it generates a proper `user` image and not a bootloader.

piotr-binkowski:
Target now generates a normal user image that works without BIOS (assigning constant values to LEDs in gateware shows proper colors) so only thing missing now is booting BIOS or other software from Flash via XIP since BIOS doesn't fit inside BRAMs.

piotr-binkowski:
I've updated the target to use BIOS stored at 0x60000 in Flash (user gateware base address + gateware size) and I've modified the platform to by default target PVT variant. During testing I also noticed an issue with yosys version that is currently used but I will create an separate issue for that.
Please take a look @mithro

PR #246, opened 2019-11-27
targets: arty: add tf target

kgugala:
Tf target is used with Zephyr Tensorflow lite demo. The target adds I2C master routed to PMOD D connector. This target is used by the Litex/Vexriscv TF lite Zephyr demo (https://github.com/antmicro/litex-vexriscv-tensorflow-lite-demo)


mithro:
FYI -- We want this on iCE40 too....

PR #248, opened 2019-11-29
use FX2Crossbar for communication with FX2 chip in HDMI2USB designs

piotr-binkowski:
This PR modifies the HDMI2USB target of Opsis and Atlys boards to utilize [FX2Crossbar](https://github.com/GlasgowEmbedded/glasgow/commits/master/software/glasgow/gateware/fx2_crossbar.py).
Using this crossbar it is possible to access 4 FX2 USB endpoints using LiteX Streams.
Current implementation attaches output from MJPEG encoder and UART to the crossbar which still leaves one free endpoint.

This depends on timvideos/HDMI2USB-fx2-firmware#47


gateware/fx2_crossbar/verilog/ddr_iobuf.v
```diff
@@ -0,0 +1,46 @@
+module ddr_iobuf(input clk, input d, input oe, output q, inout io);
```
mithro: What is the `ddr_iobuf` used for?

piotr-binkowski: It is a replacement for `SB_IO` iCE40 primitive that was used in the original crossbar design https://github.com/GlasgowEmbedded/glasgow/blob/367ecc4e83cf644c994cd25ab204e019f7568b9c/software/glasgow/gateware/fx2_crossbar.py#L289
It is used as a IO buffer for all signals connected to the FX2 chip

mithro: This should be done in migen, not in verilog?

mithro: You basically want a Spartan 6 specific version of the `_RegisteredTristate` module @ https://github.com/GlasgowEmbedded/glasgow/blob/367ecc4e83cf644c994cd25ab204e019f7568b9c/software/glasgow/gateware/fx2_crossbar.py#L267-L299

Could you also please create timing diagrams like the one that @smunaut created in https://github.com/GlasgowEmbedded/Glasgow/issues/89#issuecomment-444853359 to verify that the behaviour is correct when using the Spartan 6 IO blocks? You should also be referencing the info at https://www.xilinx.com/support/documentation/user_guides/ug381.pdf

mithro: Could you also add timing diagrams like @smunaut did at https://github.com/GlasgowEmbedded/Glasgow/issues/89#issuecomment-444853359 which verify things will work correctly with the Spartan 6? You should reference https://www.xilinx.com/support/documentation/user_guides/ug381.pdf

Is there a way to also replicate the testing that @whilequark did to prove that everything is working correctly here?

mithro:
Could you rebase this commit?

mithro:
@mateusz-holenko -- Could you work with @piotr-binkowski to get this pull request merged?

PR #297, opened 2020-01-21
Linux on Spartan6

mmicko:
Due to changes in litex (mainly due to initial values of integrated_rom_size and integrated_sram_size) I was not able to update to current master since that breaka all builds. So this contain update of litex only to version that contains my changes in ISE + Yosys flow.

Waxwing platform needed to be updated to reflect official port mapping from https://productdata.numato.com/assets/downloads/fpga/waxwing/developmentboard/WaxwingSpartan6DevelopmentBoardV3.ucf

Also added net configuration which makes ethernet work on board as well.

*NOTE:* part of code is disabled in net.py, but fix for that is available later in litex, so it can be enabled later.

To use Yosys with ISE flow just execute:
```MAKE_LITEX_EXTRA_CMDLINE="--build-option mode yosys" make gateware```

### DIFF
--- a/platforms/waxwing.py
+++ b/platforms/waxwing.py
@@ -4,8 +4,6 @@
 # Updated to:
 # https://productdata.numato.com/assets/downloads/fpga/waxwing/developmentboard/WaxwingSpartan6DevelopmentBoardV3.ucf
 #
-# Note: To use serial over FTDI change channel B to UART mode and 
-#       set driver to Virtual COM port using FTProg tool or similar
 from litex.build.generic_platform import *
 from litex.build.xilinx import XilinxPlatform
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by mithro** [comment]
FYI - I started working on modernizing the target files here -> https://github.com/mithro/litex-buildenv/commit/46aea961e195c11d484f99b1851dd5ab5b1f4770

File: platforms/waxwing.py [hunk 1]
```diff
@@ -1,6 +1,11 @@
 # Support for the Numato Waxwing Spartan 6 Development Module
 # https://numato.com/product/waxwing-spartan-6-fpga-development-board
-
+#
+# Updated to:
+# https://productdata.numato.com/assets/downloads/fpga/waxwing/developmentboard/WaxwingSpartan6DevelopmentBoardV3.ucf
+#
+# Note: To use serial over FTDI change channel B to UART mode and
```
mithro: You should probably put this comment above the USB fifo definition?

mmicko:
Great @mithro, thanks for doing modernization. That should help updating to latest litex and libraries.
Have added notes on proper places now, how to configure FTDI device, and also added emulator_ram so it can start booting linux.
Since with this version of litex there is problem applying clock on ethernet it is left unconstrained, so it does not boot still to Linux.

mithro:
This LGTM, I'm going to merge it.

PR #310, opened 2020-01-30
shadow_base was replaced by base_address

stefanor:
I *assume* this is the right change, following enjoy-digital/litepcie@000e79e11c7ac9883eaa7a10e46be9bd1deae227.

It generates, but doesn't build on the NeTV2, due to resource limits.


mithro:
@mateusz-holenko -- Could you work with @stefanor to get this pull request merged?

PR #333, opened 2020-02-12
build-zephyr: Generate csr.csv before DTS overlay

mateusz-holenko:
Make sure that `csr.csv` file is generated before using it to generate DTS overlay.


mithro:
LGTM!

PR #334, opened 2020-02-14
Added de10lite board definition file to platforms

teezzan:
I am trying to add Intel/Altera support to the repo. This will enable Fupy to be used on Intel boards such as the DE10-lite and DE0-Nano. 

I will add a base.py under the target directory. 
This is my first contribution to an open source project. Pardon my mistakes...

### DIFF
--- a/platforms/de10lite.py
+++ b/platforms/de10lite.py
@@ -1,0 +1,106 @@
+# This file is Copyright (c) 2019 msloniewski <marcin.sloniewski@gmail.com>
+# License: BSD
+
+from litex.build.generic_platform import *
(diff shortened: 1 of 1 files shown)


mithro:
Generally looks good, but the real test is to add a target and make sure everything builds.

This will be the first Altera board supported, so it's likely to be a bit more bumpy than we would like...

teezzan:
@mithro  I already added the base file and the makefile. Could you help review the code? Thank you.

mithro:
@mateusz-holenko -- Could you work with @teezzan to get this pull request merged?

PR #340, opened 2020-02-19
Allow to boot Linux from flash

mateusz-holenko:
This PR adds the definition of flash layout for Linux binaries (required for the offsets to be included in the LiteX configuration `csr.csv` file) and modifies `scripts/build-renode.sh` script so it now loads them.

NOTE:  This requires https://github.com/litex-hub/litex-renode/pull/22 to merged first!

### DIFF
--- a/scripts/build-renode.sh
+++ b/scripts/build-renode.sh
@@ -39,6 +39,7 @@ case $CPU in
 		;;
 esac
 
+LITEX_RENODE="$TOP_DIR/third_party/litex-renode"
 LITEX_CONFIG_FILE="$TARGET_BUILD_DIR/test/csr.csv"
 if [ ! -f "$LITEX_CONFIG_FILE" ]; then
 	make firmware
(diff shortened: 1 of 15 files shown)


mithro:
Looked like everything errored on Travis - can you try pushing again and see if that fixes it?

scripts/build-renode.sh
```diff
@@ -75,12 +76,30 @@ RENODE_RESC="$RENODE_SCRIPTS_DIR/litex_buildenv.resc"
...
-	--firmware-binary "$TARGET_BUILD_DIR/software/$FIRMWARE/firmware.bin" \
-	--configure-network ${TAP_INTERFACE:-""}
+
+KERNEL_FLASH_OFFSET=`python $LITEX_RENODE/litex-config-parser.py $LITEX_CONFIG_FILE --constant kernel_image_flash_offset`
+ROOTFS_FLASH_OFFSET=`python $LITEX_RENODE/litex-config-parser.py $LITEX_CONFIG_FILE --constant rootfs_image_flash_offset`
+DTB_FLASH_OFFSET=`python $LITEX_RENODE/litex-config-parser.py $LITEX_CONFIG_FILE --constant device_tree_image_flash_offset`
+EMULATOR_FLASH_OFFSET=`python $LITEX_RENODE/litex-config-parser.py $LITEX_CONFIG_FILE --constant emulator_image_flash_offset`
+
+if [ "$FIRMWARE" == "linux" ]; then
+	python $LITEX_RENODE/generate-renode-scripts.py $LITEX_CONFIG_FILE \
```
mithro: Rather than doing this parsing in bash / shell, could we change `generate-renode-scripts.py` to understand the flash offsets? IE Something like;
`--flash-binary "KERNEL_FLASH_OFFSET:$TARGET_BUILD_DIR/software/linux/firmware.bin"` and then `generate-renode-scripts.py` looks up the `KERNEL_FLASH_OFFSET` value in the csr output?

mateusz-holenko: My idea was to make `--flash-binary` switch generic -- i.e., allow to load any binary under any offset. This, however, required to know the offset beforehand - hence the need of the second parser. I agree that this probably increases the complication too much.

I agree with your suggestion - I will keep the current syntax for `--flash-binary`, i.e., `--flash-binary PATH:OFFSET`, but allow `OFFSET` to be either a number or one of predefined text labels (like `KERNEL_FLASH_OFFSET`) with values parsed from the `csr.csv` configuration file.

mateusz-holenko: I created a PR with the suggested changes to `generate-renode-scripts.py`: https://github.com/litex-hub/litex-renode/pull/23.

mateusz-holenko:
As for the failing Travis - I created a separate branch that contains only LiteX bump: https://github.com/antmicro/litex-buildenv/tree/bump_submodules.

I'll fix all the problems there and create a PR of it, then rebase this one.

scripts/build-renode.sh
mateusz-holenko: using labels parsed by `generate-renode-scripts.py` now

mateusz-holenko:
This should be rebased aftrer merging #348

mateusz-holenko:
@mithro Rebased onto the newest master - it's ready to be merged now.

targets/utils.py
```diff
@@ -139,6 +139,14 @@ def platform_toolchain_extend(platform, cmdname, argument):
     bt.clear()
     bt.extend(_platform_toolchain_cmd_join(cmds))
 
+def define_flash_constants(soc):
```
mithro, line 142: I think this should probably go into upstream LiteX or something in the future?

targets/utils.py
```diff
@@ -139,6 +139,14 @@ def platform_toolchain_extend(platform, cmdname, argument):
...
+    soc.add_constant("FLASH_BOOT_ADDRESS", soc.flash_boot_address)
+    if soc.cpu_variant == "linux":
```
mithro, line 144: BTW It would be much better for these to be ordered,

 - Emulator
 - DTS
 - Kernel
 - Root FS

Emulator + DTS could be needed for non-Linux applications / operating systems. Both should be pretty small / constant size.

The RootFS is the one which is most likely to be variable in size so should be at the one at end.

Probably should also add constants for `MAX_SIZE` here?

Issue #341, opened 2020-02-19
tftp: rootfs.cpio is zero size


pdp7:
from tutorial: [HowTo-LCA2018-FPGA-Miniconf-VexRiscv-Renode](https://github.com/timvideos/litex-buildenv/wiki/HowTo-LCA2018-FPGA-Miniconf-VexRiscv-Renode#boot-linux-via-tftp)

I am trying the section "Boot Linux via tftp".  This [gist has the full log](https://gist.github.com/pdp7/287a0a5159d1877480aaaf6e513c23b1).

Overview:
```
$ export CPU=vexriscv CPU_VARIANT=linux PLATFORM=arty TARGET=net FIRMWARE=linux
$ source ./scripts/enter-env.sh
(LX P=arty C=vexriscv.linux F=linux) $
(LX P=arty C=vexriscv.linux F=linux) $ make gateware
(LX P=arty C=vexriscv.linux F=linux) $ ./scripts/build-linux.sh
```
Here is the part with tftp:
```
(LX P=arty C=vexriscv.linux F=linux) pdp7@x1:~/dev/litex-buildenv$ make tftpd_stop
# FIXME: This is dangerous...
(LX P=arty C=vexriscv.linux F=linux) pdp7@x1:~/dev/litex-buildenv$ make tftpd_start
mkdir -p build/tftpd/
Starting atftpd
Feb 19 23:35:16 x1 atftpd[16880.139858938175552]: Advanced Trivial FTP server started (0.7)
Feb 19 23:35:16 x1 atftpd[16880.139858938175552]:   running in daemon mode on port 6069
Feb 19 23:35:16 x1 atftpd[16880.139858938175552]:   bound to IP address 192.168.100.100 only
Feb 19 23:35:16 x1 atftpd[16880.139858938175552]:   logging level: 6
Feb 19 23:35:16 x1 atftpd[16880.139858938175552]:   directory: build/tftpd//
Feb 19 23:35:16 x1 atftpd[16880.139858938175552]:   user: pdp7.pdp7
Feb 19 23:35:16 x1 atftpd[16880.139858938175552]:   log file: /dev/stdout
Feb 19 23:35:16 x1 atftpd[16880.139858938175552]:   not forcing to listen on local interfaces.
Feb 19 23:35:16 x1 atftpd[16880.139858938175552]:   server timeout: Not used
Feb 19 23:35:16 x1 atftpd[16880.139858938175552]:   tftp retry timeout: 5
[...]

PR #347, opened 2020-02-25
Verify checksums of downloaded files

mateusz-holenko:
This is to make sure that rootfs, config and DTB are correct when building the linux image.

This fixes the problem reported in https://github.com/timvideos/litex-buildenv/issues/341.

Note: there is also another commit (unrelated) that fixes builds without buildroot.

### DIFF
--- a/scripts/build-linux.sh
+++ b/scripts/build-linux.sh
@@ -30,7 +30,7 @@ fi
 eval $(make env)
 make info
 
-#set -x
+set -x
 set -e
 
 if [ "$CPU_VARIANT" != "linux" ]; then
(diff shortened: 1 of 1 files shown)


scripts/build-linux.sh
```diff
@@ -61,6 +61,7 @@ if [ ${CPU} = mor1kx ]; then
 	# To rebuild, use https://ozlabs.org/~joel/litex_or1k_defconfig
 	ROOTFS_LOCATION="https://ozlabs.org/~joel/"
 	ROOTFS=${ARCH}-rootfs.cpio.gz
+	ROOTFS_MD5="3a254683a7b6f441a4acc0d4c555230a"
```
mithro: Can you log a bug to move these into a config file? They should be part of https://github.com/timvideos/litex-buildenv/blob/master/scripts/settings.sh probably....

PR #311, opened 2020-01-30
WIP -- Modernize platforms + target files

mithro:
LiteX has change a lot since LiteX-BuildEnv was created. This pull request attempts to move towards more unification with upstream.

The major changes are;
 * Use new way to update `mem_map`.
 * Use `add_csr` + `add_interrupt` rather than previous dictionaries.
 * Use `add_wb_slave` + `add_memory_region` (cleanup usage of shadow_base and similar).

TODO Items;
 - [ ] Replace the `CRG` with new `S7PLL` and similar primitives (see https://github.com/litex-hub/litex-boards/blob/2ec6bc0bdc21b0c8afd9ca1fef0579f046c78fee/litex_boards/community/targets/mimas_a7.py#L26-L43)
 - [ ] Replace the `platforms` directory with importing from [`litex-boards`](https://github.com/litex-hub/litex-boards) and extending with the extra info.
 - [ ] Move a bunch of stuff into common functions (IE `add_spiflash`, `add_debug`, `add_linux`, etc)
 - [ ] Other stuff?

Issue #277, opened 2019-12-30
Ice40 HX8K EVB issues


niklasnisbeth:
The BIOS does not compare up on the Lattice ice40 HX8K EVB. There are several issues.

I have gotten some of the way here:

https://github.com/niklasnisbeth/litex-buildenv/commit/846fc6193972692f9de8d99146f29702d6ff80e3

I have been trying to use litex-buildenv to make a vertexriscv image for the HX8K. I am using the stub firmware. The built firmware does not come up.

We tried adding the UARTWishboneBridge, but `wishbone-tool` cannot connect to it (as if the Wishbone bus is locking up).

Moving the SRAM to address 0 makes it possible to inspect the wishbone bus with the tool.

The CPU was not running code correctly, changing the SPIFlash to little endian corrects this.

At this point, with UARTWishboneBridge still in, the CPU starts running the BIOS but is stuck in an infinite loop apparently in a call to vsnprintf in the BIOS that I cannot locate.

Removing the UARTWishboneBridge does not make the BIOS come up.

litex-buildenv revision 96ab220b0f60e3d509ca0a66dcfb81676ef294d9 results in a working Micropython environment.

niklasnisbeth:
Had a little time to dig a bit further into this, I can build a working LM32 firmware until commit da7488c2ab5e3d496ba4ecb5e482458302fb1c15.

From 3a9127d08d8d0be9a68331bd71bf43fc216dbb26, builds are broken due to memory region stuff that gets fixed in 053ec4d21a378cc65bb293c5e900ba210cd0ecf6, but at this point the LM32 BIOS does not start.

VexRiscV and Picorv are broken for me even in da7488 - I don't know if that's a separate error to the one that gets introduced in 3a912, nor if these builds were ever working on this board. I will focus on LM32 for the the time being since there's a working version to compare to.

3a912 is rather large, unfortunately. I'll attach debuggers and keep at it as time permits and make notes here.

ewenmcneill:
FTR, it seems like this same issue [affects the TinyFPGA BX](https://github.com/timvideos/litex-buildenv/issues/282) as well, which has the same iCE40 FPGA chip on it, and the same reliance on execute-in-place from the SPI flash.  I've closed https://github.com/timvideos/litex-buildenv/issues/282 in favour of consolidating the discussion here as this issue was opened first (although FTR there's some possibly useful speculation in #282 comments which might be useful for reference).

Ewen

[...]

PR #348, opened 2020-02-26
Update submodules and modernize platforms/target files

mateusz-holenko:
This updates submodules, including LiteX, to the newer versions and fixes compatibility problems.

---

*EDIT*: This also includes rebased (and a bit reworked) changes from #311.


0001-emulator-Use-external-hw-common.h-from-LiteX.patch
```diff
@@ -0,0 +1,82 @@
+From f88b259ebacb84624f2df3cfd14db56a5dd2ca4f Mon Sep 17 00:00:00 2001
```
mateusz-holenko: Patching is just a temporary solution. There is a PR to VexRiscv pending with the same changes: https://github.com/SpinalHDL/VexRiscv/pull/115.

targets/arty/base.py
```diff
@@ -162,7 +160,7 @@ def __init__(self, platform, spiflash="spiflash_1x", **kwargs):
                 endianness=self.cpu.endianness)
         self.add_constant("SPIFLASH_PAGE_SIZE", 256)
         self.add_constant("SPIFLASH_SECTOR_SIZE", 0x10000)
-        self.add_wb_slave(mem_decoder(self.mem_map["spiflash"]), self.spiflash.bus)
```
mateusz-holenko: I'm not sure about all the changes removing `mem_decoder`.

The problem with it is that the function returns lambda function instead of a number, so it's not even type compatible with `add_wb_slave`.

What was the reason of using `mem_decoder` in the first place?

scripts/build-linux.sh
```diff
@@ -166,11 +166,16 @@ if [ ${CPU} = vexriscv ]; then
 		RAM_BASE_ADDRESS=${RAM_BASE_ADDRESS::-1}
 	 	EMULATOR_RAM_BASE_ADDRESS=${EMULATOR_RAM_BASE_ADDRESS::-1}
 
+		# this is a temp fix for building the emulator
+		cd $TOP_DIR/third_party/litex/litex/soc/cores/cpu/vexriscv/verilog/ext/VexRiscv
+		git am $TOP_DIR/0001-emulator-Use-external-hw-common.h-from-LiteX.patch
```
mithro: Let's not include this at the top level.

targets/arty/base.py
mithro: https://github.com/timvideos/litex-buildenv/pull/311

mateusz-holenko:
> Can you merge this with #311 ?

Do you mean you don't want to merge this branch itself, but together with #311 OR that this will be merged first and #311 should be rebased on top of this?

scripts/build-linux.sh
mateusz-holenko: Moved to the newly-created folder `patches`.

targets/galatea/base.py
```diff
@@ -11,217 +9,57 @@
...
-        self.platform.add_period_constraint(self.crg.cd_sys.clk, 1e9/clk_freq)
-
-        # Basic peripherals
+    mem_map = {**SoCSDRAM.mem_map, **{
+        'spiflash': 0x20000000,
+    }}
+
+    def __init__(self, platform, spiflash="spiflash_1x", **kwargs):
+        dict_set_max(kwargs, 'integrated_rom_size', 0x8000)
+        dict_set_max(kwargs, 'integrated_sram_size', 0x4000)
```
mateusz-holenko: @mithro - https://github.com/timvideos/litex-buildenv/pull/311/files#diff-cbdf402788c4a014f5ad1b7253b4f18eR26 changes integrated sram size for this platform to `0x8000` - is this intended or just a typo?

targets/ice40_hx8k_b_evn/base.py
```diff
@@ -49,10 +50,13 @@ class BaseSoC(SoCCore):
     )
     csr_map_update(SoCCore.csr_map, csr_peripherals)
 
-    mem_map = {
+    mem_map_overlay = {
```
mateusz-holenko: @mithro - do you want to keep this commit? Most of the changes taken from here are reverted in the following commit anyway.

mateusz-holenko: I mean **this** commit: https://github.com/timvideos/litex-buildenv/pull/348/commits/cf423281d98ddd3ac5bda4cd225c10f71053b9ab

targets/ice40_hx8k_b_evn/base.py
```diff
@@ -80,15 +81,18 @@ def __init__(self, platform, **kwargs):
         # SoCCore ----------------------------------------------------------------------------------
         SoCCore.__init__(self, platform, sys_clk_freq, **kwargs)
 
-        self.cpu.use_external_variant("gateware/cpu/VexRiscv_Fomu_Debug.v")
+        if isinstance(self.cpu, VexRiscv):
+            self.cpu.use_external_variant("gateware/cpu/VexRiscv_Fomu_Debug.v")
```
mateusz-holenko: This does not compile on Travis at all.

@mithro - is this just a leftover of a debugging process, or should this file be included in this PR?

targets/arty/base.py
```diff
@@ -19,6 +19,7 @@
 class BaseSoC(SoCSDRAM):
     mem_map = {**SoCSDRAM.mem_map, **{
         'spiflash': 0x20000000,
+        'emulator_ram': 0xa0000000
```
mateusz-holenko: @mithro - I have no idea what the correct offset of `emulator_ram` should be, but withtout this entry in `mem_map` the platform failed on Travis.

mateusz-holenko:
Rebased onto the newest master.

Since LTO changes were reverted in LiteX, this PR now bumps LiteX to the current master.

mateusz-holenko:
@mithro Did you have time to look at my comments?

mkimage.py
```diff
@@ -8,6 +8,102 @@
...
+    Region(0x000000, 0x000100, 'Gateware', 'FPGA bitstream')
+    Region(0x000100, 0x000180, 'BIOS'    , 'LiteX BIOS with CRC')
+    Region(0x000180, 0x0003e8, 'Firmware', 'Firmware in FBI format')
+
+    """
+    region_gw = Region(
+        "Gateware",
+        "FPGA bitstream",
+        0,
+        round(size_gw),
```
mithro: Really we should be rounding upwards?

targets/arty/base.py
mithro: Should it be the same as what ever https://github.com/litex-hub/linux-on-litex-vexriscv uses?

targets/utils.py
```diff
@@ -1,27 +1,15 @@
...
-def period_ns(freq):
-    return 1e9/freq
+import difflib
+import pprint
 
 
-def csr_map_update(csr_map, csr_peripherals):
-    csr_map.update(dict((n, v)
-        for v, n in enumerate(csr_peripherals, start=(max(csr_map.values()) + 1) if csr_map else 0)))
+def dict_set_max(d, key, value):
```
mithro: This won't work if `key` doesn't exist in the dictionary.

test/common.py
```diff
@@ -4,9 +4,8 @@
 import sys
```
mithro: This looks good, but did you test the tooling works at all?

targets/upduino_v1/base.py
```diff
@@ -108,8 +103,7 @@ def __init__(self, platform, **kwargs):
 
         # Disable final deep-sleep power down so firmware words are loaded
         # onto softcore's address bus.
-        platform.toolchain.build_template[3] = "icepack -s {build_name}.txt {build_name}.bin"
-        platform.toolchain.nextpnr_build_template[2] = "icepack -s {build_name}.txt {build_name}.bin"
+        platform.toolchain.build_template[2] = "icepack -s {build_name}.txt {build_name}.bin"
```
mithro: This should call `platform_toolchain_extend` I think?

targets/tinyfpga_bx/internal.py
```diff
@@ -0,0 +1,177 @@
...
+        self.flash_boot_address = self.mem_map["spiflash"]+platform.gateware_size+bios_size+platform.bootloader_size
+
+        # Disable USB activity until we switch to a USB UART.
+        #self.comb += [platform.request("usb").pullup.eq(0)]
+
+        # Arachne-pnr is unsupported- it has trouble routing this design
+        # on this particular board reliably. That said, annotate the build
+        # template anyway just in case.
+        # Disable final deep-sleep power down so firmware words are loaded
+        # onto softcore's address bus.
```
mithro: This should call `platform_toolchain_extend` I think?

targets/tinyfpga_bx/internal.py
```diff
@@ -0,0 +1,177 @@
...
+reset_button =  [
+    # Reset shorts these two pins together
+    # Pin 12 - Silkscreen 12
+    ("reset", 0, Pins("GPIO:11"), IOStandard("LVCMOS33"), Misc("PullUp")),
+    # Pin 13 - Silkscreen 13
+    ("reset_out", 0, Pins("GPIO:12"), IOStandard("LVCMOS33")),
+]
+
+
+class _CRG(Module):
```
mithro: This should go into a `crg.py` file to be consistent with the other platforms?

targets/sim/video.py
```diff
@@ -3,7 +3,6 @@
 from litevideo.output.common import *
```
mithro: Does this sim target actually work at all? If not - add a FIXME at the top (but don't require fixing it in this pull request).

targets/arty/net.py
```diff
@@ -1,53 +1,50 @@
...
-        self.platform.add_period_constraint(self.ethphy.crg.cd_eth_tx.clk, 40.0)
+        # Ethernet MAC
+        ethmac_win_size = 0x2000
+        self.submodules.ethmac = LiteEthMAC(
+            phy        = self.ethphy,
+            dw         = 32,
+            interface  = "wishbone",
+            endianness = self.cpu.endianness)
+        self.add_wb_slave(self.mem_map["ethmac"], self.ethmac.bus, ethmac_win_size)
+        self.add_memory_region("ethmac", self.mem_map["ethmac"], ethmac_win_size, type="io")
```
mithro: There was some difference between `add_mem_region` and `add_memory_region` -- Can you check with @enjoy-digital what the right form is?

targets/galatea/base.py
mithro: Does lines 188-192 on the left indicate that was the original sizes?

targets/ice40_up5k_b_evn/base.py
```diff
@@ -122,8 +117,7 @@ def __init__(self, platform, **kwargs):
 
         # Disable final deep-sleep power down so firmware words are loaded
         # onto softcore's address bus.
-        platform.toolchain.build_template[3] = "icepack -s {build_name}.txt {build_name}.bin"
-        platform.toolchain.nextpnr_build_template[2] = "icepack -s {build_name}.txt {build_name}.bin"
+        platform.toolchain.build_template[2] = "icepack -s {build_name}.txt {build_name}.bin"
```
mithro: This should call `platform_toolchain_extend` I think?

targets/icebreaker/base.py
```diff
@@ -134,7 +129,6 @@ def __init__(self, platform, **kwargs):
 
         # Disable final deep-sleep power down so firmware words are loaded
         # onto softcore's address bus.
-        platform.toolchain.build_template[3] = "icepack -s {build_name}.txt {build_name}.bin"
-        platform.toolchain.nextpnr_build_template[2] = "icepack -s {build_name}.txt {build_name}.bin"
+        platform.toolchain.build_template[2] = "icepack -s {build_name}.txt {build_name}.bin"
```
mithro: This should call `platform_toolchain_extend` I think?

targets/icefun/base.py
```diff
@@ -102,8 +97,7 @@ def __init__(self, platform, **kwargs):
 
         # Disable final deep-sleep power down so firmware words are loaded
         # onto softcore's address bus.
-        platform.toolchain.build_template[3] = "icepack -s {build_name}.txt {build_name}.bin"
-        platform.toolchain.nextpnr_build_template[2] = "icepack -s {build_name}.txt {build_name}.bin"
+        platform.toolchain.build_template[2] = "icepack -s {build_name}.txt {build_name}.bin"
```
mithro: This should call `platform_toolchain_extend` I think?

targets/tinyfpga_bx/base.py
```diff
@@ -118,7 +113,6 @@ def __init__(self, platform, **kwargs):
         # template anyway just in case.
         # Disable final deep-sleep power down so firmware words are loaded
         # onto softcore's address bus.
-        platform.toolchain.build_template[3] = "icepack -s {build_name}.txt {build_name}.bin"
-        platform.toolchain.nextpnr_build_template[2] = "icepack -s {build_name}.txt {build_name}.bin"
+        platform.toolchain.build_template[2] = "icepack -s {build_name}.txt {build_name}.bin"
```
mithro: This should call `platform_toolchain_extend` I think?

mithro:
@mateusz-holenko -- Also take a look at https://github.com/timvideos/litex-buildenv/issues/277#issuecomment-599024874

ewenmcneill:
> @mateusz-holenko -- Also take a look at [#277 (comment)](https://github.com/timvideos/litex-buildenv/issues/277#issuecomment-599024874)

And related to that see https://github.com/timvideos/litex-buildenv/issues/366 which I opened about updating litex.  Possibly I should close #366 and we should just sort out details in this pull request?  Or the pull request should close #366 when merged?

I've done nothing further since that comment that @mithro pointed at (quoted above), because... the last couple of weeks have been busy with other things.  I'd potentially have time to test something this weekend/next week if there's something specific to test.  Now I've got TinyFPGA BX/MicroPython to boot again, it's much easier to test if other newer things still work. 

The main thing to be aware of is that I did need a fairly recent (approx 2 weeks ago) litex, because of Makefile LTO-related issues (see https://github.com/enjoy-digital/litex/issues/417 and https://github.com/enjoy-digital/litex/issues/139 for the timing of the revert of the LTO feature in litex).

For convenience, my WIP changes are in:

https://github.com/ewenmcneill/litex-buildenv/tree/refactor-tinyfpga-bx

only for TinyFPGA BX target, on top of litex-buildenv from a couple of weeks ago.  Feel free to steal anything useful from there (commits, or approach, or whatever).

Ewen

targets/tinyfpga_bx/base.py
ewenmcneill: Also IIRC the litex defaults for those changed in a useful way, so it might not even be necessary to override them any longer (just as the nextpnr override wasn't necessary any more as it's now the default).  In my WIP on the TinyFPGA BX (see https://github.com/timvideos/litex-buildenv/pull/348#issuecomment-604296404) I just removed both of them, and it seemed to be working the same as before.

(Same comment applies to several points above, but just posting once for brievity :-) )

ewenmcneill:
Also re https://github.com/timvideos/litex-buildenv/pull/348#discussion_r396855887, which GitHub doesn't have a `Reply` on any more, and `dict_set_max()` on RAM/ROM, the solution I ended up with in my TinyFPGA attempt was different and might be more flexible.

See https://github.com/ewenmcneill/litex-buildenv/commit/b788ec0a89269f7df7e0c231f84585eb62adb219.  I added an `$(TARGET_EXTRA_CMDLINE)` to the top level command line which the target `Makefile` could set, and then default that if not set in the environment.  That allows the target to have more *or* less memory than the default, which might be more useful/flexible than just always forcing the ram to a default max value.  

(Litex changing its default RAM/ROM sizes to non-zero, non-useful sizes is one of the most problematic changes in the last 6 months in terms of backwards compatibility.)

Ewen

targets/arty/base.py
mateusz-holenko: Up to this commit: https://github.com/litex-hub/linux-on-litex-vexriscv/commit/be661da7ea44969caaa62f041ffd5faed9764b14 linux-on-litex-vexriscv used `0x20000000` as `emulator_ram` base and `0xd0000000` as `spiflash` base.

Right now emulator is loaded into the main RAM under `0xc1100000`.

targets/tinyfpga_bx/base.py
mateusz-holenko: @ewenmcneill You are right. I also noticed that `-s` switch to `icepack` is now by default in LiteX. 

I verified the content of `build_template` before and after modifying it and can confirm that the above line does not change anything and is not necessary anymore. I'll remove it from all affected files.

targets/icefun/base.py
mateusz-holenko: `-s` is now passed to `icepack` in LiteX by default, so this update is not needed anymore

targets/icebreaker/base.py
mateusz-holenko: `-s` is now passed to `icepack` in LiteX by default, so this update is not needed anymore

targets/ice40_up5k_b_evn/base.py
mateusz-holenko: `-s` is now passed to `icepack` in LiteX by default, so this update is not needed anymore

targets/upduino_v1/base.py
mateusz-holenko: `-s` is now passed to `icepack` in LiteX by default, so this update is not needed anymore

mkimage.py
mateusz-holenko: That's probably right - I'll rework `round` function to ensure this.

targets/utils.py
mateusz-holenko: I'll fix this.

targets/tinyfpga_bx/internal.py
mateusz-holenko: `-s` is now passed to `icepack` in LiteX by default, so this update is not needed anymore

targets/sim/video.py
mateusz-holenko: It seems that `TARGET=sim` does not work at all:
```
(LX P=sim C=vexriscv R=update_submodules) houen@bakura:~/buildenv/348-bump_litex$ make gateware
mkdir -p build/sim_base_vexriscv/
time python -u ./make.py --platform=sim --target=base --cpu-type=vexriscv --iprange=192.168.100     \
        2>&1 | tee -a /home/houen/buildenv/348-bump_litex/build/sim_base_vexriscv//output.20200330-205713.log; (exit ${PIPESTATUS[0]})
Traceback (most recent call last):
  File "./make.py", line 169, in <module>
    main()
  File "./make.py", line 121, in main
    platform = get_platform(args)
  File "./make.py", line 51, in get_platform
    exec("from platforms.{} import Platform".format(args.platform), globals())
  File "<string>", line 1, in <module>
  File "/home/houen/buildenv/348-bump_litex/platforms/sim.py", line 1, in <module>
    from litex.boards.platforms import sim
ImportError: cannot import name 'sim' from 'litex.boards.platforms' (/home/houen//buildenv/348-bump_litex/third_party/litex/litex/boards/platforms/__init__.py)

real    0m0,155s
user    0m0,144s
sys     0m0,012s

File: targets/tinyfpga_bx/internal.py (hunk 13 above)
Comment (mateusz-holenko) [reply targets/tinyfpga_bx/internal.py]: The implementation of `_CRG` class in `targets/tinyfpga_bx/internal.py` is a bit different  than the one in `targets/tinyfpga_bx/base.py`.

It adds two reset signals, connects `user_led` and initializes special of type `SB_PLL40_CORE` - are those changes necessary and should be used in `base` target as well, @mithro? It comes from your commit https://github.com/timvideos/litex-buildenv/commit/574809ef1814acdffa23d28e0def6ceefd2201cf.

File: targets/arty/net.py (hunk 15 above)
Comment (mateusz-holenko) [reply targets/arty/net.py]: `add_mem_region` has been removed from LiteX in https://github.com/enjoy-digital/litex/commit/8bc420679a3d7dbb4c89ec5e110e4e606d088752 - now only `add_memory_region` version is available.

File: targets/galatea/base.py (hunk 4 above)
Comment (mateusz-holenko) [reply targets/galatea/base.py]: On the current master `integrated_sram_size` for `galatea` target is set to 0x4000 - https://github.com/timvideos/litex-buildenv/blob/master/targets/galatea/base.py#L192.

**Comment by mateusz-holenko** [comment]
> I added an `$(TARGET_EXTRA_CMDLINE)` to the top level command line which the target `Makefile` could set, and then default that if not set in the environment. That allows the target to have more _or_ less memory than the default, which might be more useful/flexible than just always forcing the ram to a default max value.

@ewenmcneill - I totally agree that your solution is more flexible and probably could be incorporated in all buildenv's targets, but I'm not sure if this PR is the right place to do this.

I see buildenv as a great tool to quickly start your work with LiteX by using one of many _predefined_ GW setups, HW targets and firmware to run on it. Once you are more advanced you can always dig into the boards/targets py files and tweak it to your needs. 

What I mean is that LiteX allow for a lot of customization (not only memory size) and my question is if we want buildenv to expose it all in it's own way (e.g., using Makefile args or env variables)? If so, we should probably think of some unified way of doing this and that's why I believe this deserves a separate issue/PR.

**Comment by mateusz-holenko** [comment]
> The main thing to be aware of is that I did need a fairly recent (approx 2 weeks ago) litex, because of Makefile LTO-related issues (see [enjoy-digital/litex#417](https://github.com/enjoy-digital/litex/issues/417) and [enjoy-digital/litex#139](https://github.com/enjoy-digital/litex/issues/139) for the timing of the revert of the LTO feature in litex).

Yes, I'm aware of the problems with LTO :(

> For convenience, my WIP changes are in:
> 
> https://github.com/ewenmcneill/litex-buildenv/tree/refactor-tinyfpga-bx
> 
> only for TinyFPGA BX target, on top of litex-buildenv from a couple of weeks ago. Feel free to steal anything useful from there (commits, or approach, or whatever).

Thanks! It looks nice. In fact I already borrowed one of your commits fixing region size/length.

I also encnountered problems with outdated `hw` directory in `micropython/fupy`, but https://github.com/ewenmcneill/litex-buildenv/commit/7383aa5b70f45e8b9e144140be116a9bcdd5d3af was not enough for me to make it work. I had also to (and yes, that's a hack) rename `micropython/fupy/hw` directory, because otherwise it was still used as a source of `common.h` file (and that's the outdated version of it). I'm currently looking into this.

**Comment by mateusz-holenko** [comment]
> And related to that see #366 which I opened about updating litex. Possibly I should close #366 and we should just sort out details in this pull request? Or the pull request should close #366 when merged?

TBH I would like to merge this PR as soon as possible because there are other PRs waiting for the fresh LiteX as well. In this context I suggest keeping #366 open for now and verify what's left to be done after merging this PR. What do you think @ewenmcneill ?

**Comment by mateusz-holenko** [comment]
> I also encnountered problems with outdated `hw` directory in `micropython/fupy`, but [ewenmcneill@7383aa5](https://github.com/ewenmcneill/litex-buildenv/commit/7383aa5b70f45e8b9e144140be116a9bcdd5d3af) was not enough for me to make it work. I had also to (and yes, that's a hack) rename `micropython/fupy/hw` directory, because otherwise it was still used as a source of `common.h` file (and that's the outdated version of it). I'm currently looking into this.

I ended up with the following solution: https://github.com/antmicro/micropython/commit/5ce800e32f77d6ddfeb39f308f45e8dedf1e946b, but   I'm wondering why it wasn't necessary for @ewenmcneill.

File: test/common.py (hunk 10 above)
Comment (mateusz-holenko) [reply test/common.py:4]: No, I haven't. What is the best way to run those tests?

I just notices that the following tests still have the old imports that won't work with the updated LiteX.

```test_sdram_bist.py:3:16:from litex.soc.tools.remote import RemoteClient
  test_etherbone.py:7:16:from litex.soc.tools.remote import RemoteClient
  test_alignment.py:4:16:from litex.soc.tools.remote import RemoteClient  
  test_hdmi_out.py:3:16:from litex.soc.tools.remote import RemoteClient   
  test_analyzer.py:4:16:from litex.soc.tools.remote import RemoteClient    
  test_delays.py:4:16:from litex.soc.tools.remote import RemoteClient       
  test_sdram.py:2:16:from litex.soc.tools.remote import RemoteClient       
  test_xadc.py:1:16:from litex.soc.tools.remote import RemoteClient         
  test_ddr3.py:2:16:from litex.soc.tools.remote import RemoteClient         
  test_leds.py:2:16:from litex.soc.tools.remote import RemoteClient          
  test_dump.py:2:16:from litex.soc.tools.remote import RemoteClient  
```

ewenmcneill:
> [Re setting sizes from Makefiles]
> @ewenmcneill - I totally agree that your solution is more flexible and probably could be incorporated in all buildenv's targets, but I'm not sure if this PR is the right place to do this.

@mateusz-holenko Fair enough.  I do agree it'd be worth getting this PR to a state where we're happy we can merge it, and merging it soon.

I can easily make another PR after this is merged to add that flexibility back in again. (It was there previously as the previous upstream litex default sizes were "0", which meant specifying anything was an override; whereas the change to using `max()` as the way of 'setting the default' means it's only possible to change the RAM size up from the command line.)

FWIW, I was only aiming at restoring the *same* level of flexibility that was there previously; rather than adding more flexibility in as well.  I definitely agree that *additional* flexibility is either "use upstream litex" or "we should do a larger design of how to do that".

Ewen

ewenmcneill:
> I also encnountered problems with outdated `hw` directory in `micropython/fupy`, but [ewenmcneill@7383aa5](https://github.com/ewenmcneill/litex-buildenv/commit/7383aa5b70f45e8b9e144140be116a9bcdd5d3af) was not enough for me to make it work. I had also to (and yes, that's a hack) rename `micropython/fupy/hw` directory, because otherwise it was still used as a source of `common.h` file (and that's the outdated version of it). I'm currently looking into this.

@mateusz-holenko Sorry, I didn't mention that I'd also got a WIP version of MicroPython that has a commit removing the `hw` directory, for the same reason you found.  It didn't seem relevant to this pull request, but it seems like you were testing the same "known to work previously" path as me :-) 

https://github.com/ewenmcneill/fupy-micropython/tree/modern-litex has my change, which is just the obvious one commit (https://github.com/ewenmcneill/fupy-micropython/commit/e23216bf29912cfe6dad22c9a5f42451625234e7).

Your solution (https://github.com/antmicro/micropython/commit/5ce800e32f77d6ddfeb39f308f45e8dedf1e946b; ie changing the include order) will definitely work, but will leave behind the "probably wrong for everyone" `hw` directory, and I suspect just cause confusion later.  My feeling is either we update the `hw` directory in the FuPy `ports/fupy/hw` with fresh contents from `litex` once we've merged this PR, or we just remove it entirely.  Having it there, but "not actually used" will be technical debt :-)

Finally, to answer everything in fewer messages:
> In this context I suggest keeping #366 open for now and verify what's left to be done after merging this PR. What do you think @ewenmcneill ?

I'd be inclined to do "Closes" 366 on this PR.  And make a new issue if we find there are other things to be done.  #366 was just about updating litex to a 2020 version so we weren't behind (and in particular weren't stuck on a litex version which broke non-power-of-2 RAM sizes), and that will presumably be solved by this PR.

Ewen

ewenmcneill:
@mateusz-holenko FWIW, I've got an idea how to fix the "LTO Makefile" issue in upstream Litex (I think it was really only a side effect of the structure of the change around `LDFLAGS` that impacted me; and maybe some functions needing to be marked essential that affected Florent).  It might be a while before I have focus to come back to that (my day job is... especially hectic these last couple of weeks).  But it feels like it's relatively easy to fix in upstream litex, especially once this PR is merged and it's easier to test the upstream changes in the context of litex-buildenv (which is where the Makefile / LDFLAGS issues were breaking for me).

Ewen

mateusz-holenko:
> I can easily make another PR after this is merged to add that flexibility back in again. (It was there previously as the previous upstream litex default sizes were "0", which meant specifying anything was an override; whereas the change to using `max()` as the way of 'setting the default' means it's only possible to change the RAM size up from the command line.)
> 
> FWIW, I was only aiming at restoring the _same_ level of flexibility that was there previously; rather than adding more flexibility in as well. I definitely agree that _additional_ flexibility is either "use upstream litex" or "we should do a larger design of how to do that".

You are right, using `max()` approach limits the current flexibility of buildenv by not allowing easily (i.e., using just env variables) to  _reduce_ the amount of memory. So far you could do that by setting `LITEX_EXTRA_CMDLINE` to pass additional arguments to LiteX. It required some internal knowledge on how buildenv works, but was possible (except for the fact that there were bugs in LiteX with parsing arguments de facto not allowing you to do so, but that's a different story ;)).

My biggest concern with using default value defined in the target's Makefile is that the information about the memory size is scattered accross many places: LiteX defaults, target's Makefile, target's python file and it might be hard for the user to understand what's going on and where does the configuration comes from. That's why I'm opting for a separate issue to discuss and desing the solution that is both flexible and user-friendly (and general for all targets).

> https://github.com/ewenmcneill/fupy-micropython/tree/modern-litex has my change, which is just the obvious one commit ([ewenmcneill/fupy-micropython@e23216b](https://github.com/ewenmcneill/fupy-micropython/commit/e23216bf29912cfe6dad22c9a5f42451625234e7)).
> 
> Your solution ([antmicro/micropython@5ce800e](https://github.com/antmicro/micropython/commit/5ce800e32f77d6ddfeb39f308f45e8dedf1e946b); ie changing the include order) will definitely work, but will leave behind the "probably wrong for everyone" `hw` directory, and I suspect just cause confusion later. My feeling is either we update the `hw` directory in the FuPy `ports/fupy/hw` with fresh contents from `litex` once we've merged this PR, or we just remove it entirely. Having it there, but "not actually used" will be technical debt :-)

I agree that removing the stale `hw` directory is __the__ way to go. Updating its' content would probably work in the short term, but we would find ourselves in the same situation in the future.
Keeping the code we don't want the user to use anyway is confusing and error-prone.
Silly I didn't go that way myself and started introducing even more complication in the `Makefile`.

> I'd be inclined to do "Closes" 366 on this PR. And make a new issue if we find there are other things to be done. #366 was just about updating litex to a 2020 version so we weren't behind (and in particular weren't stuck on a litex version which broke non-power-of-2 RAM sizes), and that will presumably be solved by this PR.

You are right. This PR updates LiteX so it does exactly what 366 is about. If we find any problems we can always create new issues.

ewenmcneill:
> My biggest concern with using default value defined in the target's Makefile is that the information about the memory size is scattered accross many places: LiteX defaults, target's Makefile, target's python file and it might be hard for the user to understand what's going on and where does the configuration comes from. That's why I'm opting for a separate issue to discuss and desing the solution that is both flexible and user-friendly (and general for all targets).

I  definitely agree with you, that scattering the defaults between litex (upstream) files, litex-buildenv files, target files, *and* target `Makefile`s is undesirable.  And for the same reason as you I think it's best to defer that part (ie, re-allowing RAM size overrides) to another issue.

FTR, what I was *hoping* to be able to do is detect whether "RAM size was set on the command line, or allowed to default".  And then use that to trigger whether or not the per-target default was applied.  Unfortunately there are relatively few arguments passed through to the target `__init__` so despite spending a bunch of timing trying to find a way to do that, I couldn't see an obvious non-extensive rewrite that made that possible.  (Previously this effectively worked because the upstream litex default as "0 RAM", which then allowed the per target default or the user-specified override to be obvious.)

It might be that we want to, eg, set a minimum RAM size, and use the default if the specified RAM size is less than that minimum RAM size.  (Current litex default is 4kB, which feels small enough that we could say "always more RAM than that.)

Ewen

P:S Comment here for reference; I'll then "reference in a new issue" it into a new issue so we don't forget to come back to this later, and can ignore it here.

mateusz-holenko:
@ewenmcneill  - I cherry-picked more commits from you branch. Will you have some spare time to test the current version of this branch on the HW?
You mentioned that you have an access to TinyFPGA BX, right? 

I'm currently testing it on Arty - I do that by simply running:
* the default HDMI2USB firmware on lm32,
* Micropython on vexriscv,
* Zephyr on vexrsicv
and observing if it boots correctly and responds to an input on uart.

ewenmcneill:
>  You mentioned that you have an access to TinyFPGA BX, right?

@mateusz-holenko Yes, I have a TinyFPGA BX among other devices.  I was testing with TinyFPGA BX mostly because it was convenient and tiny to use for "laptop" development :-)

Seems to me like it's working.  With the steps below to build PR #348, I get basically the same results as I was getting with my changes (compare, eg, with https://github.com/timvideos/litex-buildenv/issues/277#issuecomment-599024874).

```
git fetch origin pull/348/head:pr348
git checkout pr348
git submodule update --init --recursive
scripts/download-env.sh
make clean
make gateware
scripts/build-micropython.sh
make image-flash
```

gives me a working litex BIOS, with 10kB of RAM, and a working MicroPython.

I notice that you're temporarily pinned to my MicroPython WIP change.  Possibly we should merge that into the FuPy MicroPython if we're agreed that's the direction we want to go?  Then you could change the pin back to the FuPy one.

Ewen

```
(LX P=tinyfpga_bx.minimal F=micropython R=pr348) ewen@parthenon:/src/fpga/litex-
buildenv$ make firmware-connect
flterm --port=/dev/ttyUSB0 --speed=115200
[FLTERM] v2.4-29-g47d3b15 Starting...

        __   _ __      _  __
       / /  (_) /____ | |/_/
      / /__/ / __/ -_)>  <
     /____/_/\__/\__/_/|_|
   Build your hardware, easily!

 (c) Copyright 2012-2020 Enjoy-Digital
 (c) Copyright 2007-2015 M-Labs

 BIOS built on Apr  6 2020 09:27:48
 BIOS CRC passed (931d7dc5)

 Migen git sha1: 3f9809b
 LiteX git sha1: 536ae0e6

--=============== SoC ==================--
CPU:       LM32 @ 16MHz
ROM:       32KB
SRAM:      10KB

--============== Boot ==================--
Booting from serial...
Press Q or ESC to abort boot completely.
sL5DdSMmkekro
Timeout
Booting from flash...
Executing booted program at 0x20058008

--============= Liftoff! ===============--
MicroPython v1.9.4-1430-ge23216bf2 on 2020-04-06; litex with lm32
>>> print("Hello World!")
Hello World!
>>> 
```

```
(LX P=tinyfpga_bx.minimal F=micropython R=pr348) ewen@parthenon:/src/fpga/litex-
buildenv$ git log -2
commit 421a402a68824dda81395c58675c1d55c21a0261 (HEAD -> pr348)
Author: Mateusz Holenko <mholenko@antmicro.com>
Date:   Fri Apr 3 11:02:18 2020 +0200

    micropython: temporarily change the fork
    
    The new fork containt a commit removing an outdated `hw`
    directory generated by LiteX.

commit d9d9cf24393e5ea39808d25fb33a693ebdc7cbe4
Author: Ewen McNeill <ewen@naos.co.nz>
Date:   Fri Mar 13 22:19:22 2020 +1300

    micropython: link to litex hw includes
    
    (Instead of relying on third_party/ports/fupy having a cached copy of the
    "hw" includes, which could get out of date; and was out of date after updating
    litex; modern litex has a different set of CSR routines.)
(LX P=tinyfpga_bx.minimal F=micropython R=pr348) ewen@parthenon:/src/fpga/litex-
buildenv$ 
```

mithro:
@ewenmcneill - Any idea if the ROM is coming from SPI flash? (I guess we should also put the ROM source in the BIOS output?)

ewenmcneill:
> @ewenmcneill - Any idea if the ROM is coming from SPI flash? (I guess we should also put the ROM source in the BIOS output?)

@mithro I can't see how the TinyFPGA `rom` segment wouldn't be coming from the SPI flash, given https://github.com/timvideos/litex-buildenv/blob/421a402a68824dda81395c58675c1d55c21a0261/targets/tinyfpga_bx/base.py#L36-L39 and https://github.com/timvideos/litex-buildenv/blob/421a402a68824dda81395c58675c1d55c21a0261/targets/tinyfpga_bx/base.py#L68-L73

It was definitely coming from the SPI Flash in the version I was working on, because I looked at the linker maps.

Ewen

mithro:
@ewenmcneill - Great! Just wanted to check. @mateusz-holenko Is this ready to merge now then?

mateusz-holenko:
@mithro I tested it on Arty and both the default HDMI2USB and Micropython boot fine.
I was planning to test it today also on Nexys Video and netv2.

Otherwise I believe it's ready to be merged.

mateusz-holenko:
> I notice that you're temporarily pinned to my MicroPython WIP change. Possibly we should merge that into the FuPy MicroPython if we're agreed that's the direction we want to go? Then you could change the pin back to the FuPy one.

@ewenmcneill That was exactly my idea. Could you create a PR to FuPy MicroPython? Once it's merged I will point back to the original repository.

mateusz-holenko:
@mithro Travis is red, but it's caused by some network problems when downloading conda packages:
```
('Connection broken: OSError("(104, \'ECONNRESET\')")', OSError("(104, 'ECONNRESET')"))
```
It happens a lot lately :(

All other tests passed.

mateusz-holenko:
Tested on `nexys_video` and `netv2` - the default firmware boots nicely to prompt.

The only problem was that for `nexys_video` I had to disable `CAS` module as there were errors during synthesis because `user_btn5` and `cpu_reset` signals are using the same pin.

Targets tested on HW:

- [x] arty
- [x] nexys_video
- [x] netv2
- [x] tinyfpga_bx

mithro:
Great! Let's merge for now and continue with more pull requests.

ewenmcneill:
> @ewenmcneill That was exactly my idea. Could you create a PR to FuPy MicroPython? Once it's merged I will point back to the original repository.

https://github.com/fupy/micropython/pull/58, which I've just merged since I have commit rights on that repo, and since we all seem in agreement with the approach.

We'll want to revert https://github.com/timvideos/litex-buildenv/commit/421a402a68824dda81395c58675c1d55c21a0261 (which changes it to use my fork/branch) fairly soon.  @mateusz-holenko Do you want do do that revert?  Or should I create a revert commit/PR?  (The two are currently the same, other than the merge commit in the FuPy MicroPython version.)

Ewen

mateusz-holenko:
@mithro Great!

@ewenmcneill I created a PR: https://github.com/timvideos/litex-buildenv/pull/385

PR #398, opened 2020-04-14
pano_logic_g2: Set RTL serial port baudrate from environment BAUD if …

skiphansen:
…present.

### DIFF
--- a/targets/pano_logic_g2/base.py
+++ b/targets/pano_logic_g2/base.py
@@ -16,6 +16,7 @@ from gateware import spi_flash
 from .crg import _CRG
 
 from targets.utils import dict_set_max
+import os
 
 class BaseSoC(SoCSDRAM):
     mem_map = {**SoCSDRAM.mem_map, **{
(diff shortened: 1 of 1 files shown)


targets/pano_logic_g2/base.py
```diff
@@ -26,6 +27,10 @@ class BaseSoC(SoCSDRAM):
     def __init__(self, platform, **kwargs):
         dict_set_max(kwargs, 'integrated_rom_size', 0x8000)
         dict_set_max(kwargs, 'integrated_sram_size', 0x8000)
+        if os.environ.get('BAUD', '0') == '0':
+            kwargs['uart_baudrate']=19200
+        else:
+            kwargs['uart_baudrate']=int(os.environ.get('BAUD'))
```
mithro, line 33: Maybe something like the following;
```python
        baud = os.environ.get('BAUD', '19200')
        try:
            kwargs['uart_baudrate'] = int(baud)
        except TypeError as e:
            raise ArgumentError('BAUD value in environment {} is not a valid number!'.format(baud)
```

I also wonder if this should be part of the `make.py` script to enable it for all designs?

mithro: I guess it should probably check that there isn't a `uart_baudrate` already in kwargs?

mithro:
@skiphansen - I think the idea of this change makes sense and I support adding it -- just a couple of random comments around the implementation.

targets/pano_logic_g2/base.py
skiphansen: makes sense to me.  I was trying to limit my damage to Pano.  I'll give that a try.  Thanks for the review!

skiphansen:
I'll be submitting a new patch that's more general.

PR #399, opened 2020-04-16
Add support for setting --target-option from TARGET_OPTIONS env var

skiphansen:
I successfully tested this by setting the uart_baudrate target option on the pano_logic_g2 platform.  In theory the other options should work as well, but I didn't test them.

(LX P=pano_logic_g2 C=vexriscv.linux F=linux) skip@dell-790:~/pano/working/litex-buildenv$ echo $TARGET_OPTIONS
uart_baudrate=961538

### DIFF
--- a/make.py
+++ b/make.py
@@ -52,9 +52,44 @@ def get_platform(args):
     return Platform(**dict(args.platform_option))
 
 
+def get_env_target_options():
+    int_options = ['uart_baudrate','integrated_rom_size',\
+                   'integrated_sram_size','cpu_reset_address',\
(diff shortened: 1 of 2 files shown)


skiphansen:
Yea, it got a bit out of hand.   I'll see what I can do.  (Frankly this is my first nontrivial attempt at Python, I'm a C programmer!)

skiphansen:
I tried adding the values to sys.argv but that doesn't work expect for string options.

Ewen pointed out the existing --uart-baudrate command line switch so this change wasn't really needed in the first place.  Closing PR.

PR #409, opened 2020-04-18
Pano: Add support for specifying alternate UART connection

skiphansen:
Several Pano hackers use the DDC lines on the HDMI port for a serial connection rather than the DVI port.  This change allows the connection to be HDMI port to be used without code modifications.


mithro:
Awesome! 

Might not hurt to create a wiki page which describes how to connect a UART to each of these ports (ideally with some pictures).

skiphansen:
I've updated https://github.com/timvideos/litex-buildenv/wiki/HowTo-Linux-on-Pano-Logic-G2.  

I don't have any pictures currently as the way I made my cable was a royal pain and I am not recommending that others repeat the same mistake.  I'll ask the Pano community and see if anyone else had pictures they can share.

Issue #386, opened 2020-04-08
Choose a better order for the flash image / offsets


mithro:
BTW It would be much better for these to be ordered,

 - Emulator
 - DTS
 - Kernel
 - Root FS

Emulator + DTS could be needed for non-Linux applications / operating systems. Both should be pretty small / constant size.

The RootFS is the one which is most likely to be variable in size so should be at the one at end.

Probably should also add constants for `MAX_SIZE` here?

_Originally posted by @mithro in https://github.com/_render_node/MDExOlB1bGxSZXF1ZXN0Mzc3MTI2MDA2/pull_requests/unread_timeline_

PR #410, opened 2020-04-19
Panologic: Added SPI flash support.  Successfully boots Linux from flash.

skiphansen:


mithro:
LGTM!

mithro:
@mateusz-holenko - Can you take a quick look and merge if it seems fine?

targets/pano_logic_g2/base.py
```diff
@@ -73,7 +73,32 @@ def __init__(self, platform, **kwargs):
...
+        self.submodules.spiflash = spi_flash.SpiFlash(
+            platform.request("spiflash"),
+            dummy=platform.spiflash_read_dummy_bits,
+            div=platform.spiflash_clock_div,
+            endianness=self.cpu.endianness)
+        self.add_csr("spiflash")
+        self.add_constant("SPIFLASH_PAGE_SIZE", platform.spiflash_page_size)
+        self.add_constant("SPIFLASH_SECTOR_SIZE", platform.spiflash_sector_size)
+        self.add_constant("SPIFLASH_TOTAL_SIZE", platform.spiflash_total_size)
+        self.add_wb_slave(
```
mateusz-holenko, line 86: Maybe instead of using `add_wb_slave` and `add_memory_region` you could just call `register_mem` just like in https://github.com/timvideos/litex-buildenv/blob/master/targets/icefun/base.py#L78?

targets/pano_logic_g2/base.py
```diff
@@ -73,7 +73,32 @@ def __init__(self, platform, **kwargs):
...
+            self.mem_map["spiflash"],
+            self.spiflash.bus,
+            platform.spiflash_total_size)
+        self.add_memory_region(
+            "spiflash",
+            self.mem_map["spiflash"],
+            platform.spiflash_total_size)
+        self.flash_boot_address = self.mem_map["spiflash"]+platform.gateware_size
+        self.add_constant("FLASH_BOOT_ADDRESS", self.flash_boot_address)
```
mateusz-holenko, line 95: Are the image offsets of DT/emulator/kernel/rootfs in any way special and fixed (apart from the fact that it must fit in the flash ofc)?

I'm asking because there is `define_flash_constants` helper function with a similar logic (https://github.com/timvideos/litex-buildenv/blob/master/targets/utils.py#L142) that could potentially be used here. It won't work right now, because the `gateware_size` is big and the offsets used there would go out of flash, but it should be usable after fixing https://github.com/timvideos/litex-buildenv/issues/386.

For now this version is fine, but I'll like to refactor it in the future to use similar layout for all targets. What do you think about it?

skiphansen: Yes the offsets are special in that  I wanted to put the dts, kernel and root filesystem at flash erase page boundaries so they could be flashed independently.  I'm planning on doing some Linux application development and expect to be updating the rootfs frequently w/o changes to the bit file, kernel or dts.  It takes a LONG time to flash the entire image!  I'm planning on adding support to the target specific make file for flashing.

An addition complication is the LX100 base "Rev C" Pano which has a smaller flash.  Linux will not fit so the offsets will (eventually) need to be conditioned on the device.

mateusz-holenko: Sure - let's leave it as it is then!

PR #413, opened 2020-04-24
Fix path to vivado on Travis

mateusz-holenko:
Paths in XILINX_BINDIR are already absolute so
prepending them with XILINX_BINDIR is wrong.

It worked locally because XILINX_DIR is only set
when XILINX_PASSPHRASE is present (which is the case
of the CI setup).


mateusz-holenko:
Do you mean https://github.com/timvideos/litex-buildenv/blob/master/.travis/build.sh? I don't see any `build.sh` in gateware directory.
If so, it could be put in `download-xilinx.sh` script - there is a very similar fragment of code to the one in `enter-env.sh`.

I placed updating PATH in `enter-env.sh`, because it's needed not only for tests in CI, but also when using buildenv normally on a local machine. Since we look there for the toolchain in `/opt/Xilinx` anyway I thought we could update PATH as well. Of course we can always change the approach and expect user to add `vivado` to PATH and in the `enter-env.sh` script just ensure it's available.

mithro:
@mateusz-holenko https://github.com/timvideos/HDMI2USB-firmware-prebuilt/blob/master/archive/master/v0.0.4-584-g2011661/arty/net/vexriscv.linux/gateware/build_top.sh

mithro:
@mateusz-holenko The file generated by LiteX in the gateware directory.

mateusz-holenko:
https://github.com/enjoy-digital/litex/commit/0b923aa49757e777302ce8d28e35414977fc9db0 removes the support for automatic sourcing from LiteX - now vendor tools are required to be in PATH.

mateusz-holenko:
@mithro Do you want to modify the `build_top.sh` script (generated by LiteX and calling just `vivado`) in buildenv by setting a full path to the binary?

mateusz-holenko:
@mithro According to the discussion in https://github.com/enjoy-digital/litex/issues/495 I reworked the proposed solution. It now monkey-patches functions generating toolchain scripts in LiteX by injecting Xilinx settings file sourcing right before running vivado/ise.

mateusz-holenko:
@mithro What do you think of this approach? It required to copy-paster and modify the content of `litex.build.xilinx.vivado._build_script()` and `litex.build.xilinx.ise._run_ise()` functions from LiteX.

For Vivado it would be possible not to copy the whole function from LiteX (although it's pretty short anyway), but just wrap the original one and modify the generated `build_*.sh` script afterwards.

This won't work, however, for ISE as the `_run_ise` not only generates the `build_*.sh` script, but also executes it right away.

mateusz-holenko:
@enjoy-digital implemented an alternative solution in https://github.com/enjoy-digital/litex/commit/b4267a79011fa7dd7cb84961196875d3be81f077

mateusz-holenko:
@mithro Yes, LiteX provides an option to set an environment variable with the path to Vivado/ISE which is a much less intrusive solution. I'm this solution with: https://github.com/antmicro/litex-buildenv/commit/c230f74674e49745f00806a0456f08d747d13341

PR #414, opened 2020-04-24
Add support for CPU_VARIANT to load.py.

skiphansen:


skiphansen:
Is load.py used by anyone?  I tripped across it while I was adding loading/flashing support for the Pano, but other than my uncommitted local code I can't can find any references to it in the source tree or the Wiki. 

Maybe it should just be deleted?

mithro:
@skiphansen - You seem to be right in that nothing currently uses `load.py` -- I think we might want more things to use `load.py` in the future however? It could be extended in the future to understand more advance loading modes like `tftp`?

PR #432, opened 2020-05-13
Optionally allow checkout commits and BR2_EXTERNAL to be specified.

skiphansen:
Recent commits to linux-on-litex-vexriscv for SMP broke Linux for the Pano Logic platform.  These changes allow the linux-on-litex-vexriscv and the buildroot repositories to be checked out from specific (known working) commits.  

If BD_COMMIT, LLV_COMMIT and BR2_EXTERNAL are not set in the environment then the script behavior is unchanged.

Allowing BR2_EXTERNAL to be specified in the environment makes it possible to use a customized buildroot configuration for a specific platform.  For example to enable DHCP on eth0.

See https://github.com/skiphansen/panog2_linux/blob/master/Makefile for an example of how this can be used.

### DIFF
--- a/scripts/build-linux.sh
+++ b/scripts/build-linux.sh
@@ -235,6 +235,10 @@ if [ ${CPU} = vexriscv ] && [ ${BUILD_BUILDROOT:-0} = 1 ]; then
 			cd $(dirname $BD_SRC)
 			echo "Downloading Buildroot code."
 			git clone $BD_REMOTE $BD_SRC
+			cd $BD_SRC
+			if [ x$BD_COMMIT != x ]; then
+			    git checkout $BD_COMMIT
+			fi
 		)
 		fi
 
@@ -243,6 +247,9 @@ if [ ${CPU} = vexriscv ] && [ ${BUILD_BUILDROOT:-0} = 1 ]; then
 			cd $(dirname $LLV_SRC)
 			echo "Downloading Linux on LiteX-VexRiscv code."
 			git clone $LLV_REMOTE $LLV_SRC
+			if [ x$LLV_COMMIT != x ]; then
+			    (cd $LLV_SRC; git checkout $LLV_COMMIT)
+			fi
 		)
 		fi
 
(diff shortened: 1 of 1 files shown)


scripts/build-linux.sh
```diff
@@ -258,7 +265,11 @@ if [ ${CPU} = vexriscv ] && [ ${BUILD_BUILDROOT:-0} = 1 ]; then
 		dtc -I dts -O dtb -o $TARGET_LINUX_BUILD_DIR/rv32.dtb $TARGET_LINUX_BUILD_DIR/rv32.dts
 
 		cd $BD_SRC
-		make BR2_EXTERNAL=$LLV_SRC/buildroot/ litex_vexriscv_defconfig
+                if [ "x$BR2_EXTERNAL" = "x" ]; then
```
mithro, line 268: I'm a bit suspicious of this if statement. Can you check it @mateusz-holenko?

mateusz-holenko, line 268: there is something wrong with the indent + I guess `""` is not needed if you use the x trick

mateusz-holenko: I checked it locally and it seems that having a `BR2_EXTERNAL` variable set in the shell is not enough. I had to export it using `export BR2_EXTERNAL` in order for `make` to see/use it.

This test do not check if the variable is exported, so I guess it might give wrong results.

skiphansen: Thanks, I'll fix it.  My editor showed it correctly, it probably has a different tab setting or something.

scripts/build-linux.sh
```diff
@@ -235,6 +235,10 @@ if [ ${CPU} = vexriscv ] && [ ${BUILD_BUILDROOT:-0} = 1 ]; then
 			cd $(dirname $BD_SRC)
 			echo "Downloading Buildroot code."
 			git clone $BD_REMOTE $BD_SRC
+			cd $BD_SRC
```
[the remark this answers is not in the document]
skiphansen: I'll move it.

scripts/build-linux.sh
skiphansen: Good point, I have only tested with BR2_EXTERNAL exported, I'll fix it.

mateusz-holenko: The current solution should work, but wouldn't it be simpler to use:
```
make BR2_EXTERNAL=${BR2_EXTERNAL:-$LLV_SRC/buildroot} litex_vexriscv_defconfig
```
?

skiphansen: Yes, but I think that has the same issue that you commented on yesterday.  Namely it doesn't test if BR2_EXTERNAL has been exported or simply set.  Personally I'm not too worried about that.

mateusz-holenko: I mean, just `make ...` without any `ifs` around it.

With this apporach you don't need to export `BR2_EXTERNAL` in your shell anymore. AFAIK `make X=1` is not the same as `X=1 make`.

I tested it locally and it seemed to work.

skiphansen: Done!

mateusz-holenko: Looks good.

PR #450, opened 2020-06-02
linux: unify linux source

mateusz-holenko:
This commit changes the linux source
for VexRiscv and mor1kx to litex-hub/linux.

Since this is the same repository as used
by linux-on-litex-vexriscv and all the LiteX
drivers are included, there is no more need for
additional BUILD_BUILDROOT option.

This also updates rootfs images for both architectures
and adds buildroot configs used to generate them.


mateusz-holenko:
Now that the LiteX submodule is updated I'll rebase this and remove the draft status.

PR #479, opened 2020-07-06
Add support for the internal TFTP server in Renode

mateusz-holenko:
This switches to the internal TFTP server in Renode as a default binaries source for net-enabled platforms.


scripts/build-renode.sh
```diff
@@ -58,17 +58,48 @@ if [ ! -z "$ETH_BASE_ADDRESS" ]; then
...
 		make tftp
 
-		TAP_INTERFACE=tap0
+		RENODE_CONFIG="--configure-network tap0"
+		;;
+
+	internal)
+		echo "Using the Renode internal TFTP server for netbooting"
+
+		if [ "$FIRMWARE" = "linux" ] && [ "$CPU" = "vexriscv" ]; then
```
mithro, line 67: Should this be part of the config generation rather than passing this all via the command line?

scripts/build-renode.sh
```diff
@@ -58,17 +58,48 @@ if [ ! -z "$ETH_BASE_ADDRESS" ]; then
...
+		fi
+
+		RENODE_CONFIG="$RENODE_CONFIG
+			--tftp-server-ip \"192.168.100.100\"
+			--tftp-server-port 6069"
 		;;
 
 	none)
-		echo "Renode networking disabled..."
+    	        # this will be handled later
```
mithro, line 83: Currently this sounds like a "TODO: Add handling here" rather than "This part of the case statement is handled separately at line 90." 

```suggestion
    	        # This case is handled a bit further below.
```

scripts/build-renode.sh
mateusz-holenko: This is the part of the resc/repl generation - those parameters are passed to `generate-renode-scripts.py`, not Renode itself. 

I think having `generate-renode-scripts` as generic as possible is better - that way we can support many different `linux` configurations (different file names, different offsets) without the need of changing the generating script itself.

scripts/build-renode.sh
mateusz-holenko: You are right, I rephrased it as suggested.

PR #493, opened 2020-07-21
linux: mor1kx: Add support for the external DT

mateusz-holenko:
This updates submodules (including the LiteX submodule) and restores the support for the external DT file for linux on mor1kx.


mithro:
Merge if you are happy with this!

PR #502, opened 2020-08-02
Fix typo for CAS rgb_led in a couple platforms (arty, cmod_a7).

tcal-x:
These platforms used "rgb_leds", but the routine in gateware/cas.py uses "rgb_led", so no RGB LED IOs were being generated for these platforms.   This fixes that.

Signed-off-by: Tim Callahan <tcal@google.com>


tcal-x:
I noticed that the generated RGB outputs are only driven correctly for RGB LED0, but not for RGB LEDs 1, 2, and 3.   I'll try to fix that.

Issue #503, opened 2020-08-05
Current yosys version fails on some iCE40 designs


piotr-binkowski:
While working on #244 I noticed that some designs result in yosys failing on `yosys-abc` call:
```
4.46.18. Finished OPT passes. (There is nothing left to do.)

4.47. Executing TECHMAP pass (map to technology primitives).

4.47.1. Executing Verilog-2005 frontend: /home/pbinkowski/work/litex-buildenv/litex-buildenv/build/conda/bin/../share/yosys/ice40/latches_map.v
Parsing Verilog input from `/home/pbinkowski/work/litex-buildenv/litex-buildenv/build/conda/bin/../share/yosys/ice40/latches_map.v' to AST representation.
Generating RTLIL representation for module `\$_DLATCH_N_'.
Generating RTLIL representation for module `\$_DLATCH_P_'.
Successfully finished Verilog frontend.

4.47.2. Continuing TECHMAP pass.
No more expansions possible.

4.48. Executing ABC pass (technology mapping using ABC).

4.48.1. Extracting gate netlist of module `\fomu' to `<abc-temp-dir>/input.blif'..
Extracted 6754 gates and 8474 wires to a netlist network with 1718 inputs and 1147 outputs.

4.48.1.1. Executing ABC.
ERROR: ABC: execution of command "/home/pbinkowski/work/litex-buildenv/litex-buildenv/build/conda/bin/yosys-abc -s -f /tmp/yosys-abc-nuE16o/abc.script 2>&1" failed: return code 134.
```
When running the command manually I get the following output:
```
(LX P=fomu C=picorv32.minimal F=none R=fomu) Singularity> /home/pbinkowski/work/litex-buildenv/litex-buildenv/build/conda/bin/yosys-abc -s -f /tmp/yosys-abc-nuE16o/abc.script
ABC command line: "source /tmp/yosys-abc-nuE16o/abc.script".

+ read_blif /tmp/yosys-abc-nuE16o/input.blif 
+ read_lut /tmp/yosys-abc-nuE16o/lutdefs.txt 
+ strash 
+ ifraig 
+ scorr 
Warning: The network is combinational (run "fraig" or "fraig_sweep").
+ dc2 
+ dretime 
+ strash 
+ dch -f 
+ if 
+ mfs2 
+ lutpack -S 1 
[...]

PR #509, opened 2020-08-11
Add dynamic DT generation for Linux

mateusz-holenko:
This uses the `json2dts.py` script provided by LiteX to generate the DT file matching the current platform.


mateusz-holenko:
I fixed the problem with booting Linux in Renode - it turned out to be related to the way the `dtc` tool was called (apparently there is a difference in automatic detection of input/output types between versions).

Travis is still red (one of the tests fails), but it's related to the #503. Otherwise it's ready to be merged @mithro.

mateusz-holenko:
@mithro I've just realized a little problem - the DT generation script in LiteX originates from the linux-on-litex-vexriscv repo and currently supports only the VexRiscv CPU. This configuration works fine both in Renode and on HW.
Unfortunately, it doesn't work for mor1kx. I'll have to fix this in LiteX first, before merging this PR.

mateusz-holenko:
@mithro - it's fixed now. Both VexRiscv and Mork1x work on HW + VexRiscv works in Renode. This is ready to be merged.

PR #586, opened 2020-12-02
litex-renode: Update submodule

mateusz-holenko:
This introduces support for creating Renode scripts/platform from the csr.json file and adds generation of LiteX SPI controller node.


mithro:
Guess we need to move this repo to Github Actions?

mateusz-holenko:
> Guess we need to move this repo to Github Actions?

I'll take a look at that.

PR #614, opened 2021-01-03
ICE40 Feather support

whatnick:
- [ ] Support pinmaps for LED
- [ ] Support UART
- [ ] Support SPI/Quad-SPI
- [ ] Support using "icefeather" as a platform string
- [ ] Verify with physical board the lm32 and vexriscv micropython startup
Signed-off-by: Tisham Dhar <whatnickd@gmail.com>

### DIFF
--- a/platforms/ice40_feather.py
+++ b/platforms/ice40_feather.py
@@ -1,0 +1,73 @@
+from litex.build.generic_platform import *
+from litex.build.lattice import LatticePlatform
+from litex.build.lattice.programmer import IceStormProgrammer
+
+
+_io = [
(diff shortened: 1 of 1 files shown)


whatnick:
This is the WIP for ice-feather support. Progress so far is
- IceFeather can be made to work with soft-cpu + micropython by hacking ice-breaker definition with some pin changes
- IceFeather is recognized as a platform using an environment variable. However the gateware + firmware images built do not boot into a working micropython.

Making this up as I go along, would benefit from a adding new board HOW-TO. The boards is very very similar to icebreaker.

